PR #8, opened 2017-06-08
Update version of NPM and packages

pspeter3:
This is just to make sure the code is using up to date packages before a conversion to TypeScript.


jbucaran:
@pspeter3 I think we don't need to include the lock file for the same reasons given here:

- https://github.com/hyperapp/hyperapp/pull/223

pspeter3:
Interesting, I would have made the opposite argument. Since we have no runtime dependencies, we don't need to worry about lock files affecting downstream customers. On top of that, having a lock file should increase determinism and reduce the number of issues where developers are having inconstancy locally. Up to you though.

jbucaran:
If we need it, we can always add it later right?

pspeter3:
Yep, sounds good.

pspeter3:
I've updated with your requests

.gitignore
```diff
@@ -1,6 +1,11 @@
-yarn.lock
-example*
-.vscode
-yarn.lock
+# Artifacts
```
jbucaran: @pspeter3 No yarn.lock? I may still be using yarn.

package.json
```diff
@@ -18,12 +18,12 @@
   ],
   "scripts": {
     "test": "exit 0",
-    "build": "rollup -cm -n picodom -f umd -i index.js -o dist/picodom.js",
-    "format": "prettier --semi false --write index.js"
+    "build": "rollup --config --format umd --input index.js --name picodom --output dist/picodom.js --sourcemap",
```
jbucaran: @pspeter3 Can you go back to using the short form of the --flags? 🙏

pspeter3: My thought was that the longer form was easier to read. I can just scrap this PR.

jbucaran:
Thanks @pspeter3! 👍

PR #15, opened 2017-06-29
modify h.js for more understandable code

jabez128:
use flatten function and object spread to make h.js code more standardable : )

### DIFF
--- a/h.js
+++ b/h.js
@@ -5,9 +5,8 @@ export function h(tag, data, ...stack) {
   var children = stack.reduce((p,n)=>{
     if(n != null && n !== true && n !== false) {
           if(typeof n === 'number') n = n + '';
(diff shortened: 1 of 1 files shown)


jbucaran:
@jabez128 Thanks! I still haven't added any tests, so this change may break something! 😞 

Maybe you can help porting some of the tests from HyperApp? The catch is I don't want to use jest, instead I'd like to see how hard would it be to get the same functionality with just tape. 

I'd consider other test frameworks as well!

jabez128:
@jbucaran ok, i'll try add some unit test with tape test framework : )

jbucaran:
@jabez128 We are 100% test covered now, but I am still going to close here. I feel `h` is already as minimal and simple as it can be.

PR #17, opened 2017-07-06
Add basic DTS

### LINKED ISSUE #7: TypeScript Conversion
This looks really great. How would you feel about converting to TypeScript? (happy to help)

pspeter3:
This DTS file does not support the JSX namespace because that is a more complicated issue.

Closes #7


ngryman:
In #7 you speak about tests, do you still want to add them?

pspeter3:
Given the DTS size, I think that you're right and tests are not needed at this point.

jbucaran:
> @pspeter3 One issue though is it seems like you can not have a component that returns another component when using h. Thoughts?

Do you mean?

```jsx
const AnotherComponent = (props, children) => h("div", {}, "hi")
```
```jsx
const Component = (props, children) => AnotherComponent
```

or

```jsx
const Component = (props, children) => AnotherComponent(props, children)
```

pspeter3:
I mean the middle one where the component returns another function. I'm not sure a world which that would happen and the implementation of `h` does not handle it so I didn't make the DTS handle it. I'm not particularly worried about it.

jbucaran:
@pspeter3 Got it. Note that the following does work:

```jsx
const Component = (props, children) => h(AnotherComponent, props, children)
```

pspeter3:
@jbucaran do we still want this?

jbucaran:
@pspeter3 Definitely! 

For reference: https://github.com/hyperapp/hyperapp/pull/279

pspeter3:
So what's next to have this merged?

package.json
```diff
@@ -19,11 +19,12 @@
...
-    "format": "prettier --semi false --write *.js"
+    "format": "prettier --semi false --write *.js *.d.ts"
   },
   "devDependencies": {
     "prettier": "~1.4.2",
     "rollup": "^0.42.0",
     "rollup-plugin-uglify": "^2.0.1"
-  }
+  },
+  "typings": "picodom.d.ts"
```
jbucaran: I thought this was outdated? 

/cc @Swizz

jbucaran:
@pspeter3 Please fix the conflict first! 🙏👍

package.json
Swizz: No, not this one, the one you requested about the `typings.json` file

pspeter3:
Fixed

pspeter3:
Ping?

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    tag: string
+    data?: D
+    children: VNode<{}>
+  }
+
+  export interface Component<D> {
+    (data?: D, ...children: VNode<{}>[]): VNode<D>
+  }
+
+  export type VElement<D> = Component<D> | string
```
jbucaran: @pspeter3 Can we somehow avoid declaring VElement and just put it in the definition of `tag` below?

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  export function h<D>(
+    tag: VElement<D>,
+    data?: D,
+    ...children: VNode<{}>[]
```
jbucaran: @pspeter3 Does this mean variable Vnode args or array of Vnodes?

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
...
+  ): VNode<D>
+
+  export function patch(
+    parent: Element,
```
jbucaran: @pspeter3 Is Element the same as HTMLElement?

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    element: Element,
```
jbucaran: element can be null.

picodom.d.ts
andrewiggins: Element is base type for HTMLElement and SVGElement

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
+declare namespace picodom {
+  export interface VNode<D> {
+    tag: string
+    data?: D
+    children: VNode<{}>
+  }
+
+  export interface Component<D> {
+    (data?: D, ...children: VNode<{}>[]): VNode<D>
```
andrewiggins: Does this map to the last line in h.js: `tag(data, children)`? If so, should this be 
```typescript
export type VNodeChildren = VNode<{}> | string;

export interface Component<D> {
    (data?: D, children: VNodeChildren[]): VNode<D>
}
```

Bonus, you can reuse `VNodeChildren` in the `VNode` interface.

picodom.d.ts
```diff
@@ -0,0 +1,30 @@
+declare namespace picodom {
```
andrewiggins: Thoughts on we following the [module.d.ts UMD example](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) from the [TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#umd)?

Basically would just be removing the two declares and adding `export as namespace picodom`;

picodom.d.ts
andrewiggins: Similar to [Optional Parameters in Callbacks](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks), I don't think data needs to be optional here. The Component interface will always be invoked with a `data` argument and `children` argument. If a consumer wants to ignore them that is valid.

I tested the following locally and it passes:
```typescript
export interface Component<D> {
  (data: D, children: VNode<{}>[]): VNode<D>
}

let Test: Component<any> = () => <div></div>;
```

package.json
```diff
@@ -35,5 +35,6 @@
     "prettier": "~1.5.3",
     "rollup": "^0.45.2",
     "rollup-plugin-uglify": "^2.0.1"
-  }
+  },
+  "typings": "picodom.d.ts"
```
andrewiggins: Should this be types ([doc](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html))?

andrewiggins:
For the comments I left, I'd be happy to jump in a contribute some fixes. I'd also be happy to add some JSDoc comments to the definitions for better intellisense in editors such as VSCode

andrewiggins:
At some point (either this PR or another) we should look at exposing the available lifecycle methods through the type definitions

picodom.d.ts
jbucaran: This is correct, data is a required argument. It can be `null` or `{}` if you want to ignore it.

jbucaran: @pspeter3 By the way, I'd prefer to rename this argument to `props` instead.

jbucaran:
> @andrewiggins At some point (either this PR or another) we should look at exposing the available lifecycle methods through the type definitions.

Sounds good to me.

picodom.d.ts
pspeter3: Yes but why? It's been helpful with the React type definitions to have this type exist. I actually think it needs to be applied to more places.

picodom.d.ts
pspeter3: That works, I haven't done it in a while. Will update to that.

picodom.d.ts
```diff
@@ -0,0 +1,26 @@
+export interface VNode<D> {
+  tag: string
+  data?: D
+  children: VElement<{}>[]
```
jbucaran: @pspeter3 I think we might not need VElement, but please correct me if I am wrong. The signature of a Component is good, but a virtual node children property is always an array. We make sure of that inside `h`.

pspeter3: Yeah it's an array of `VNode | string` right? How else do you represent text nodes?

jbucaran: See what I mean:

<pre>
h("div", {}, "Foo") => { 
  tag: "div", 
  data: {}, 
  children: <b>["Foo"]</b> 
}
</pre>

jbucaran: Children is always an array. If you pass a string to h(), it will return Array<string>.

pspeter3: So VElement just means `VNode | string` which is what you have there. It's an array of them which is also what you have there.

jbucaran:
@pspeter3 Is this merge ready? and how can I test if this is working, like, how do I tell vscode to pick the typings? 

Sorry, total typings n00b here.

pspeter3:
Yeah, I think we can merge it.

jbucaran:
@pspeter3 Any tips on how to tell vscode to pick this up?

picodom.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function h<D>(
+  tag: Component<D> | string,
+  data?: D,
+  ...children: VElement<{}>[]
+): VNode<D>
+
+export function patch(
+  parent: Element,
+  element: Element,
```
jbucaran: @pspeter3 element can be null.

picodom.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface Component<D> {
+  (data?: D, ...children: VElement<{}>[]): VNode<D>
+}
+
+export type VElement<D> = VNode<D> | string
+
+export function h<D>(
+  tag: Component<D> | string,
+  data?: D,
```
jbucaran: Can we rename `D` to something more meaningful?

- Data
- DataAttributes
- VNodeData
- VNodeAttributes

andrewiggins: * Props?

pspeter3: I think Data makes sense since I took it from the function signature.

picodom.d.ts
pspeter3: Oh, good to know.

picodom.d.ts
```diff
@@ -0,0 +1,26 @@
+export interface VNode<Data> {
+  tag: string
+  data?: Data
+  children: VElement<{}>[]
+}
+
+export interface Component<Data> {
+  (data?: D, ...children: VElement<{}>[]): VNode<D>
```
jbucaran: D

pspeter3: I wish you had the ability to edit my PR. Good catch though

jbucaran:
Thank you @pspeter3 for your incredible patience with this! 🎉

Swizz:
I am coming too late, **but** I am reworking the Hyperapp DTS largely based on the picodom DTS (because you did on awesome job !)

So, `h()` also accept as parameter a children that will be an array of Vnode, so, why not overloads `h()` with : 
```tsx
export function h<Data>(
  tag: Component<Data> | string,
  data?: Data,
  children?: VElement<{}>[]
): VNode<Data>
```
Plus, `h()` is ready to be use with JSX, so why not type JSX support too ?

```tsx
declare global {
  namespace JSX {
    interface Element<Data> extends VNode<Data> {}
    interface IntrinsicElements {
      [elemName: string]: {}
    }
  }
}
```

I am sorry if this comment could be judged as pretty annoying.

andrewiggins:
@Swizz I'm working on improving the types a bit, including what you mentioned. PR should be out soon.

PR #19, opened 2017-07-24
Improve TypeScript types and add TypeScript compilation tests

### LINKED ISSUE #34: Add JSX support for Typescript
With the right declarations in the d.ts file, we should be able to compile directly to h() calls from Typescript, without an additional Babel transform after compiling - as well as getting type-safety and IDE support for JSX expressions in Typescript.

I found clues [here](https://github.com/Microsoft/TypeScript/issues/13402#issuecomment-323623043) and supposedly Snabbdom has this feature.

I will likely experiment with this myself soon, and will post results here if I make any progress :-)

andrewiggins:
I've added some sample TypeScript files to validate the typings in the repo.

* add JSDoc comments
* add lifecycle methods and key property

I've taken the feedback from @Swizz [PR in hyperapp](https://github.com/hyperapp/hyperapp/pull/311) and incorporated it here. Let me know if I've missed anything.

I'll be adding comments to the PR about the reason for each change.

Resolves #34, #35


.gitignore
```diff
@@ -3,6 +3,7 @@ dist
 example*
 .gz
 coverage
+test/typings/**/*.js
```
andrewiggins: I'm ignoring the generated files from the TypeScript tests. There is an option in TypeScript to no emit any files (the `noEmit` option). However, I find it helpful sometimes to peek at the generated JavaScript to understand what it compiles to, which is particularly useful for JSX.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
-export interface VNode<Data> {
+export as namespace picodom
```
andrewiggins: Since this line names the entire module, I moved it to the top.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
+
+export interface VirtualNode<Data> {
   tag: string
-  data?: Data
+  data: Data
```
andrewiggins: Per [line 26 in h.js](https://github.com/picodom/picodom/blob/901a0c8a403244cfc8ec0d143aca036b208ba491/src/h.js#L26), data will never be undefined.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
   children: VElement<{}>[]
 }
 
 export interface Component<Data> {
-  (data?: Data, ...children: VElement<{}>[]): VNode<Data>
+  (data: Data, children: VElement<{}>[]): VirtualNode<any>
```
andrewiggins: Per [line 29 in h.js](https://github.com/picodom/picodom/blob/901a0c8a403244cfc8ec0d143aca036b208ba491/src/h.js#L29), children is an array of VirtualNodes or strings. Each child is not its own argument to the Component function

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
 }
 
-export type VElement<Data> = VNode<Data> | string
+export type VElement<Data> = VirtualNode<Data> | string
+
+export function h<Data>(
+  tag: Component<Data> | string,
+  data?: Data,
+  children?: Array<VElement<{}> | number>
```
andrewiggins: `h` accepts an array of children. This enables line 7 in h-test.ts in this PR. I've also added number as an acceptable type to `h` to enable line 14 of h-test.ts.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
-export interface VNode<Data> {
+export as namespace picodom
+
+export interface VirtualNode<Data> {
```
andrewiggins: Rename `VNode` to `VirtualNode` per [PR 311 comment in hyperapp](https://github.com/hyperapp/hyperapp/pull/311)

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
+  data?: Data,
+  children?: Array<VElement<{}> | number>
+): VirtualNode<Data>
 
 export function h<Data>(
   tag: Component<Data> | string,
   data?: Data,
-  ...children: VElement<{}>[]
-): VNode<Data>
+  ...children: Array<VElement<{}> | number>
```
andrewiggins: Let's keep the discussion of whether or not to keep the argument spread version of h in [hyperapp PR 311](https://github.com/hyperapp/hyperapp/pull/311) where it is currently happening. I left it here since it already existed.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
-  oldNode: VNode<{}> | null,
-  node: VNode<{}>
+  oldNode: VirtualNode<{}> | null,
+  node: VirtualNode<{}>
 ): Element
 
-export as namespace picodom
+declare global {
+  namespace JSX {
+    interface Element<Data=any> extends VirtualNode<Data> {}
```
andrewiggins: I gave the generic argument `Data` a default type of `any` to allow people to use it as `JSX.Element` like in line 3 in jsx-test.ts, which other libraries (preact, react) allow.

picodom.d.ts
```diff
@@ -1,26 +1,41 @@
...
+    interface IntrinsicElements {
+      [elemName: string]: any
```
andrewiggins: It is important that the properties of the `IntrinsicElements` interface return a type of `any`. If it were `{}`, then line 46 of jsx-test.ts would fail because `{ id: string }` doesn't match `{}`. Instead, any allows any attribute to be set on intrinsic elements.

test/h.test.js
```diff
@@ -83,6 +83,7 @@ test("skip null and Boolean children", () => {
 
 test("components", () => {
   const Component = (data, children) => h("div", data, children)
+  const Wrapper = (data, children) => h("div", data, children.map(vn => h(Component, null, vn)))
```
andrewiggins: I've added this component and the corresponding tests to assert the type of the children argument. The implementation of h guarantees that children will always be any array and this test validates that. I discovered this because the typing of the `Component` interface was wrong.

test/typings/h-test.ts
```diff
@@ -0,0 +1,56 @@
+import { h, patch, Component } from "../../picodom";
+
+// empty vnode
```
andrewiggins: For this file I copied many of the assertions from the unit tests for `h` to validate that the typings cover all tested uses of `h`

picodom.d.ts
Swizz: Nice catch, I will use this too

andrewiggins:
Let me know if you'd like me to squash my commits, etc.

andrewiggins:
@jbucaran I've squashed my commits into one. Let me know if there is anything else you'd like me to do.

jbucaran:
@andrewiggins Can you remind me why are these files necessary? 🤔

jbucaran:
@andrewiggins Can you simplify this PR a bit? For example, if you can remove the JSX tests that would be great.

andrewiggins:
Would you mind if I removed the jsx-test.ts file and added 1 or 2 jsx tests into the h-test.ts file? Just to validate that the `h` function typings satisfy TypeScript's expectations for JSX.

andrewiggins:
In my latest push I did what I suggest in my previous comment (remove jsx + add a small number of jsx tests to h-test). Let me know what you think.

jbucaran:
@andrewiggins There are comments inside some of the test files. Also, did you run prettier on the code?

andrewiggins:
@jbucaran, I ran prettier on the code which cleaned up some rogue semi-colons I accidentally slipped in there :)

The commented lines in the test code are for lines of TypeScript code that should throw an error when compiling. I'd like to keep them in there so that future authors who update the typings file know not only what TypeScript should allow, but also when TypeScript should throw compile errors. I would need to spend more time setting up test infrastructure to do this automatically. I thought for the time being, a simple comment that says "The following line should throw an error" would suffice. If you feel strongly though, I can remove the commented lines. Just let me know.

Which files would you like to be in `src/`? I would think that the `h-test.tsx` would remain in the `test` folder since it isn't code that consumers should execute, however perhaps it makes sense to move `picodom.d.ts` into `src/` similar to what [preact](https://github.com/developit/preact/blob/master/src/preact.d.ts) does. Let me know what your thoughts are.

Thanks!

jbucaran:
@andrewiggins Thanks for your work this.

> I would think that the h-test.tsx would remain in the test folder since it isn't code that consumers should execute, however perhaps it makes sense to move picodom.d.ts into src/ similar to what preact does

Yes, I agree with both suggestions. Let's move `picodom.d.ts` inside `src/`.

And I am not sure about the naming in `h-test`, is that a TypeScript convention perhaps? I feel it should be `h.test.tsx`, but that's because I am used to that in JS land.

---

>  I would need to spend more time setting up test infrastructure to do this automatically.

I would prefer to merge this when such infrastructure is setup if that's alright with you.

andrewiggins:
> And I am not sure about the naming in h-test, is it a TypeScript convention perhaps? I feel it should be h.test.tsx, but that's because of how I write test files in JS land.

Oh, good point. I don't know why I did dashes there lol. `h.test.tsx` sounds good to me, and is consistent with the naming of the other tests. I'll fix that.

> I would strongly prefer to merge this when such infrastructure is setup if that's alright with you.

That's fine. Will take me some time to get that setup but I'll comment back here when I do :)

JorgeBucaran:
@Swizz @andrewiggins Hi guys! Now that 1.0.0 has landed and the API is stable, can we rebase and update this PR?

andrewiggins:
I'll try to get to it this weekend

JorgeBucaran:
Thanks @andrewiggins! 👍

picodom.d.ts
```diff
@@ -1,23 +1,38 @@
...
   oldNode: VNode<{}> | null,
   newNode: VNode<{}>,
-  container: HTMLElement = document.body
+  container?: HTMLElement
 ): Element
 
-export as namespace picodom
+declare global {
+  namespace JSX {
+    interface Element extends VNode<any> {}
```
andrewiggins: Note: I explicitly don't make Element generic as doing that causes TypeScript to no longer enforce JSX prop types.

picodom.d.ts
```diff
@@ -1,23 +1,38 @@
...
   type: string
   props?: Props
-  children: VNode<Props> | string
+  children: Array<VNode<{}> | string>
 }
 
-export interface Component<Props> {
-  (props?: Props, ...children: VNode<Props> | string): VNode<Props>
+export interface Component<Props = {}> {
+  (props: Props, children: Array<VNode<{}> | string>): VNode<Props>
```
andrewiggins: Instead of making props optional, I defined the default type in the generic type param since a Component function should never received `undefined` as props (re: [line 23 of h.js](https://github.com/picodom/picodom/blob/d0c95dd50678d6df220e8e642e5b7365e622bc2b/src/h.js#L23))

mindplay-dk:
@andrewiggins I tried it out, the type declarations look pretty solid :-)

I have one issue though - I try really hard to minimize my toolchain, which means my strong preference is to use the Typescript compiler only, and avoid any post-processing with Babel or otherwise.

I was unable to get this working with `outFile`, since it only works with `module` set to either `amd` or `system` - and I need to use `amd` either way, because that's what we use in-house.

But neither of those settings appear to work with these type declarations. With `module` set to either `amd` or `system`, my import statement fails with `cannot find module 'picodom'`.

The only working `module` setting appears to be `commonjs`.

Is that because of how the package itself is structured?

If so, that might be one reason to use Typescript - so that different versions of the package can be built from a single source while generating the type declarations for each type of module? /CC @JorgeBucaran

JorgeBucaran:
@mindplay-dk The source is using ES modules, and the bundle is UMD. I think you could use picodom as a module directly in the browser though.

Related: https://github.com/hyperapp/hyperapp/pull/441

mindplay-dk:
Here's my [sandbox project](https://bitbucket.org/mindplaydk/picodom-sandbox/src/?at=sans-fusebox) with just Picodom and Typescript. It doesn't work, and it may be impossible to finish this without something like FuseBox or RollUp to stitch things together? I'm not sure if it *would* work, but the fork doesn't seem to include the browser-build of Picodom... (?)

Anyways, so far so good. I'm really excited to see it compiling with type-checks and all! :-D

mindplay-dk:
Okay, so I did finally get this to build and run using FuseBox - [it works](https://bitbucket.org/mindplaydk/picodom-sandbox/src/?at=pico-fusebox) :-)

I'm extremely happy to have gotten this far, and can already see myself using pico for, well, everything ;-)

I'm still not really happy having to install FuseBox just for this - it's a ton of dependencies, and I have no real understanding of what it's doing or how. My strong preference would be getting this to work with just the Typescript compiler, but maybe I'm chasing rainbows here. This is certainly good enough that I could see myself moving ahead with this for real projects.

You guys are awesome!

andrewiggins:
You aren't chasing rainbows. This package and types should definitely work without a bundler, using only TypeScript. Checkout [this gist using picodom](https://gist.github.com/andrewiggins/a1ca8f6b344e2b36f127a52c82846d30) for a sample similar to your sandbox that uses TypeScript's `outFile` and picodom without a bundler (see the Readme for an important step to get this working). I think you are correct - including the project from github directly doesn't include the built files for the browser. You need to get the package from npm for that.

Once this PR is merged then the npm package will contain the correct typings to use the npm package out of the box, and you'll no longer need the "hack" mentioned in the readme of the gist.

andrewiggins:
@JorgeBucaran Anything else you'd like to see so we can close this? If you'd like, you should be able to [squash my commits in the github UI](https://github.com/blog/2141-squash-your-commits) when merging this PR.

Previously we talked about setting up error checking tests for TypeScript. Given that this PR contains some crucial fixes to get the definitions working, would you be willing to take this PR without them? If you feel strongly about waiting until those tests are in, I can continue working on them (work in progress happening in [this branch in this file](https://github.com/andrewiggins/picodom/blob/typings-error-test/test/ts.test.js), but its still a little ways out from being ready)

JorgeBucaran:
@andrewiggins Let's merge here first! Thanks for working on this! 🎉

package.json
```diff
@@ -32,7 +35,8 @@
     "jest": "^21.2.1",
     "prettier": "~1.7.4",
     "rollup": "^0.50.0",
+    "typescript": "^2.5.3",
     "uglify-js": "^2.7.5"
   },
-  "typings": "picodom.d.ts"
+  "typings": "./src/picodom.d.ts"
```
JorgeBucaran: @andrewiggins Can we move picodom.d.ts to the root of the project?

mindplay-dk:
@andrewiggins with target=es5, the compiler will default to module=commonjs, and with that will default to moduleResolution=node, so the latter setting is likely redundant?

I wonder if module=amd with outside set would work with moduleResolution=node though, since the default for module=amd is moduleResolution=classic though? Maybe that's why my attempt failed. I can't spot any other difference from your gist. Will try this in the morning :-)

package.json
andrewiggins: @JorgeBucaran reverted my change and added "picodom.d.ts" to the package.json `files` list so it is included in the npm package.

JorgeBucaran:
Thanks @andrewiggins! Are we ready to merge here?

andrewiggins:
@JorgeBucaran I think so! Would you like to submit an approval first? (new to this - is there a generally accepted practice?)

picodom.d.ts
```diff
@@ -1,23 +1,38 @@
...
   props?: Props,
-  ...children: VNode<Props> | string
+  ...children: Array<VNode<{}> | string | number | null>
+): VNode<Props>
+
+export function h<Props>(
+  tag: Component<Props> | string,
+  data?: Props,
+  children?: Array<VNode<{}> | string | number | null>
 ): VNode<Props>
```
JorgeBucaran: @andrewiggins Why is this repeated?

```js
export function h<Props>(
  type: Component<Props> | string,
  props?: Props,
  ...children: Array<VNode<{}> | string | number | null>
): VNode<Props>

export function h<Props>(
  tag: Component<Props> | string,
  data?: Props,
  children?: Array<VNode<{}> | string | number | null>
): VNode<Props>
```

mindplay-dk: Overloading. One accepts spread arguments `...children` and the other accepts an array.

JorgeBucaran: Right, would it be possible to express that in the same export declaration?

mindplay-dk: Type/tag and data/props could be consistently named in those declarations though. (and throughout the file.)

JorgeBucaran: Alternatively, if that is not possible, could we remove the declaration that takes spread arguments without TypeScript complaining and leave it in the source just as a detail?

mindplay-dk: I don't think you can put those in a single declaration, no - they're different call signatures because they have different number of arguments. (Typescript validates overloads for compatibility though, so the duplication is not a liability.)

mindplay-dk: I don't think you should remove the spread declaration, since that's the one that Typescript output will actually use.

JorgeBucaran: Gotcha! 👍 

Now, people using `h()` should always express children using an array. The spread arguments syntax is supported only because JSX needs it, but this may change as JSX evolves / improves.

JorgeBucaran: @mindplay-dk I thought the order was: TypeScript -> JSX -> JavaScript

mindplay-dk:
Not sure what you mean by "order"? The TSX processing is done internally by the Typescript compiler.

JorgeBucaran:
> @mindplay-dk I don't think you should remove the spread declaration, since that's the one that Typescript output will actually use.

Sure. My point is, do you mean we "should" keep it or are we required to keep it otherwise tsc complains/fails?

mindplay-dk:
I would keep it, since it's an accurate description of the function calls supported by the actual implementation - by removing it, you create an artificial limitation for Typescript users. I don't know whether the compiler needs it or not.

JorgeBucaran:
@mindplay-dk Thanks! 

> I don't know whether the compiler needs it or not.

This is what I need to find out. If the compiler doesn't care, I'd prefer to remove it, since the only reason it's in the source code is because of a JSX issue.

- https://github.com/babel/babel/issues/2034

mindplay-dk:
> If the compiler doesn't care, I'd prefer to remove it, since the only reason it's in the source code is JSX compat.

@JorgeBucaran understood, so considering Typescript doesn't use arrays, and the React and Snabbdom declarations both use `...children` exclusively - using spread arguments appears to be the norm, although the compiler doesn't seem to complain one way or the other.

To avoid lots of packing and unpacking of arrays when implementing component patterns etc. it might be nice to retain support for both though? Of course you can use the spread operator easily, but it'll compile to lots of packing/unpacking when targeting ES5.

I'd recommend retaining both, but would otherwise lean towards the spread argument form, since it seems to be more consistent with what the rest of the community are doing.

mindplay-dk:
@JorgeBucaran althoooough..... having skimmed the comments in that Babel issue, now I'm not sure - the performance impact of spread arguments is perhaps deeper than I assumed.

It's an odd situation, because Typescript accepts the factory function even though it's declared with children as an array rather than spread - yet, the compiled output is still a spread. I guess we'd have to talk to the Typescript team about changing the compiled output? Otherwise, there's no way to get the performance edge - except of course by disabling react processing and post-processing the JSX with Babel, but...

I don't know, I think the safest approach at this time is to support and type-hint for for both forms.

JorgeBucaran:
@andrewiggins @mindplay-dk [picodom@1.0.1 published](https://www.npmjs.com/package/picodom)! Thank you Rasmus for pushing this forward and testing things out and Andre for the work! 🎉🌮🎉 

screen shot 2017-11-03 at 22 04 06

PR #22, opened 2017-07-24
Add bundlesize config

andrewiggins:
Adds bundlesize config to the package.json


jbucaran:
@andrewiggins Nice, thanks! Let's be bold and make it 1.2 KB. 💪

PR #24, opened 2017-08-14
*WIP*: Flow

andyrj:
I am not 100% about the type VirtualComponent (I had missed that in my first draft for hyperapp as well), and need to test this but should be working for users to be able to use flow with picodom.

Changes, comments, suggestions always welcome.


andyrj:
so here is a commit that adds to the dist directory a copy of the *.js.flow files for flow to pick up after an npm install picodom ...  I am going to test this out locally but should be working, it has 90.5% flow coverage.  I had to add two helper functions to work around flow errors, if anyone knows of a better way to get flow to be happy I would be glad to change whatever in this PR.  I'm still learning about flow myself so this may not be the best way to do things.

I didn't like having to add the typeof conditionals but couldn't find a way to get flow to like the implicit type checks that were in the code base.

andyrj:
I can't seem to get the tests to run with babel-jest, anyone had experience with getting jest to work with es6 code?  or know what the es5 syntax would be for the line:
```js
import type { VirtualNode } from "./h"
```

jbucaran:
@andyrj It seems this PR is adding flow annotations to the source code, but the one [on hyperapp](https://github.com/hyperapp/hyperapp/pull/323) is only adding type declarations.

Can we just add type declarations?

andyrj:
Yes, that is what my first set of commits were in this PR, but I thought I would put types on the entire code base and clear up flow errors, so now I can be more sure of the standalone type declarations.  

For instance I had missed Node, which is what document.createElement() returns, needed in some of params for patch() that I found by doing this.

If you would like I can either open a new PR or revert this PR to just the picodom.js.flow and a simple build step to copy it into dist/

jbucaran:
I'd prefer keeping and fixing this one PR if you can do that.

jbucaran:
By the way, how does one test this?

andyrj:
Well I guess you could add flow types to the tests and then run flow on that directory, if you have 100% test coverage that should fully exercise the flow library definition also.

andyrj:
ok last commit worked around h() type declaration...  it now works but is not very strict on type checks with ...children in h params.  Maybe someone smarter than myself can figure out a more strict type declaration for that function

Issue #26, opened 2017-09-03
(feature) Something like `context` in React


acstll:
Would it be possible to have something like `context` in React?

An object that you could set once on initialisation (in a custom `render` function at the top of your app) and then somehow access it later down in the component tree?

It'd be mostly for convenience, since I'm already passing an object (I think you call it `data`) down, and get my state from it.

I'd like to build something similar to `react-router` to be used with this library.

What do you think? is this even possible? a stupid idea?

jbucaran:
> @acstll What do you think? is this even possible? a stupid idea?

Cool idea. Picodom, unlike hyperapp is only the patch and vnode factory, so I believe you should be able to accomplish almost anything you want by building on top of those two functions.

> An object that you could set once on initialisation (in a custom render function at the top of your app) and then somehow access it later down in the component tree?

Sounds like you already laid out some of the implementation details. How hard could it be to write this render function? 🤔😄

acstll:
After thinking about it for a bit, I believe I'm talking about a custom `h` function.

```js
let context

export function setContext (x) {
  context = x
}
export function h () {
  // ...
}
```

This `h` function would keep a reference to this "context" object and either merge it everytime into `data`, [here](https://github.com/picodom/picodom/blob/master/src/h.js#L28), or pass it as a third argument to components, [here](https://github.com/picodom/picodom/blob/master/src/h.js#L31).

[As a side note, I'm alreay using `picodom` successfully in a tiny app as part of a bigger website. I'm liking it a lot]

jbucaran:
> @acstll  I believe I'm talking about a custom h function.

 Okay, that definitely sounds like a [functor or HOF](https://www.wikiwand.com/en/Higher-order_function). I am not an expert on the terminology, so I had to look that up to make sure, but my understanding is that you would create a function that takes `h` as an argument and returns a new `h'` function that has the built-in technology you need.

acstll:
Ok, HOF sounds familiar to me because of React ([HOC](https://facebook.github.io/react/docs/higher-order-components.html)), but functor does not, haha (I'm no expert either). I think that's one way to do it. I'm going to try that out and let you know how it goes. (Thanks!)

jbucaran:
@acstll Thanks for bringing this up! I'll leave this open for a while.

@ngryman @zaceno Do you have any feedback?

zaceno:
@jbucaran @acstll I think it sounds like a very cool idea, and I look forward to seeing it in practice! 🙏

[...]

PR #27, opened 2017-09-08
Proof of concept for `context` idea

acstll:
This is an implementation of that discussed in here https://github.com/picodom/picodom/issues/26

It's very simple. Changes are:

- exports a new function called `setContext`
- all Components receive a third argument `(data, children, context)`

This third argument will be `null` in case `setContext` was never called, or it will be whatever you passed to `setContext`.

Tests have been added to keep coverage at 100%.

The use case is the same as explained in the [React documentation for context](https://facebook.github.io/react/docs/context.html):

> In some cases, you want to pass data through the component tree without having to pass the props down manually at every level.

Of course, this here is waaaaay more simple than context is in React.

If you like this, we can discuss a bit the naming (I don't really like `context` and would prefer something else) and merge it, since it's a very tiny change.

Otherwise, there's also the option of making it a separate package, like `picodom-context`, which would only export `{ h, setContext }`.

What do you think?

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -1,5 +1,10 @@
 var i
 var stack = []
+var context = null
+
+export function setContext(x) {
+  context = x
+}
 
 export function h(tag, data) {
   var node
@@ -28,5 +33,5 @@ export function h(tag, data) {
         data: data || {},
         children: children
       }
-    : tag(data, children)
+    : tag(data, children, context)
 }
--- a/src/index.js
+++ b/src/index.js
@@ -1,2 +1,2 @@
-export { h } from "./h"
+export { h, setContext } from "./h"
 export { default as patch } from "./patch"
--- a/test/h.test.js
+++ b/test/h.test.js
@@ -1,4 +1,4 @@
-import { h } from "../src"
+import { h, setContext } from "../src"
 
 test("empty vnode", () => {
   expect(h("div")).toEqual({
@@ -98,6 +98,30 @@ test("components", () => {
         tag: "div",
         data: { id: "bar" },
         children: []
+      }
+    ]
+  })
+})
+
+test("components with context", () => {
+  const Component = (data, children, context) => {
+    return context != null ?
+      h(context.foo, data, children) :
+      h("div", data, children)
+  }
+
+  expect(h((data, children, context) => context)).toEqual(null)
+
+  setContext({ foo: "p" })
+
+  expect(h("div", {}, [h(Component, {}, "baz")])).toEqual({
+    tag: "div",
+    data: {},
+    children: [
+      {
+        tag: "p",
+        data: {},
+        children: ["baz"]
       }
     ]
   })


jbucaran:
@acstll Thank you for this excellent POC. 

I'd prefer a separate package for the following reasons:

- **Simplicity**: We can't just add everything on top of picodom, but we can this time. This feature can perfectly live outside core.
- **Maintenance**: Without this feature there are less tests and code to maintain in the main repo.
- **Flexibility**: "Of course, this here is waaaaay more simple than context is in React." As a third party library, different authors can add just the right amount of "way" they want.

jbucaran:
@acstll On the other hand, this is a very small addition, so let's discuss. I can change my mind.

An official package is another option too! 🤔

jbucaran:
Comments? 🙏 @zaceno @andyrj @ngryman

acstll:
You're very welcome. My opinion is also divided. It's such a small addition it wouldn't hurt, but it also makes sense as a separate package, so you use it only if you really need it.

Yep, comments would be highly appreciated.

zaceno:
I think, unless we're sure this is something most users of picodom will want, this would be better as a separate package. That doesn't mean I'm against *ever* adding it. I think we need to see some real world usage first.

WOuld it be possible to have a working demonstration of how this is similar, but simpler than, context in React? 🙏

src/h.js
```diff
@@ -1,5 +1,10 @@
 var i
 var stack = []
+var context = null
```
jbucaran: @acstll This means all users of `h` share the same context. Is that desirable?

acstll: I think it is, that's the idea. Having access to very common stuff. When you say "users" you mean any module (internal or external) that imports `h`, right?

acstll:
> I think we need to see some real world usage first.

@zaceno I pretty much agree with this. I'm in favor of a separate package. If it turns out to be a good idea, we can evaluate adding it to core, but only then. I can create the package, and it could be later added to the picodom organization if you feel that's the right thing.

Also, being that `context` is a shared object (see https://github.com/picodom/picodom/pull/27#pullrequestreview-61438265), it makes a lot of sense for it to be separate.

Do you have any ideas for a better name, instead of `context`?

> WOuld it be possible to have a working demonstration of how this is similar, but simpler than, context in React? 

it's simpler because you don't need to declare anything, you just get it as an argument (like in Preact). It's the same because it's just an object you can use, not related to props or state (in picodom there's no such thing as state).

I "needed" this because I want to try and build something like (but smaller) `react-router` 4, which is routing using components. I can try and do that, so it'll do as a working demontration which is also useful.

zaceno:
@acstll 

> I want to try and build something like (but smaller) react-router 4, which is routing using components. I can try and do that, so it'll do as a working demontration which is also useful.

Please do - that'd be great! I haven't used React myself, so I only have a very vague idea of what "context" is in that context. So I'm very curious to see how this construct would be used in practice.

I'm imagining things like you could attach the call to rerender the dom onto the context, perhaps, or to rerender certain parts of the vdom only. Or state management solutions. My imagination runs so far ahead of me that I need someone to show me some nuts and bolts :)

src/h.js
jbucaran: Yes, that's what I meant.

jbucaran:
Another issue about adding this to core that I didn't address [here](https://github.com/picodom/picodom/pull/27#issuecomment-328019824) is that it would create a precedent: where do we draw the line of what's core and what's not core?

jbucaran:
> @acstll I can create the package, and it could be later added to the picodom organization if you feel that's the right thing. 

Sounds good to me.

> Do you have any ideas for a better name, instead of context?

Here are some ideas:

- `context`: Easier to remember since it's based on the same concept as [React's context](https://facebook.github.io/react/docs/context.html).
- `sharedProps`
- `wire`
- `channel`
- `medium`

acstll:
> I can create the package

I'll do that then, thanks a lot!

acstll:
Feel free to close this :)

jbucaran:
Let's do it then! 💪

jbucaran:
P.S: What did you decide to call it?

acstll:
> What did you decide to call it?

I think I'll go with `context`, if you come from React or Preact you know exactly what it means.

jbucaran:
@acstll Sounds good.

ngryman:
@jbucaran Hey, so sorry for the flatline heartbeat lately, so much stuff on my side 😜.

@acstll I think this is a cool proposal but it shouldn't be part of the core. `picodom` is too low-level for this. The notion of components does not really exist here, it's an additional layer left to the user.

PR #32, opened 2017-10-10
rename tag to type

zaceno:
Because hyperapp did it, so we should do it to so that vdom decorators that wish to be compatible with hyperapp and picodom can be.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -1,7 +1,7 @@
 var i
 var stack = []
 
-export function h(tag, props) {
+export function h(type, props) {
   var node
   var children = []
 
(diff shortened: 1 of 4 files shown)


src/h.js
```diff
@@ -22,11 +22,11 @@ export function h(tag, props) {
...
-  return typeof tag === "string"
+  return typeof type === "string"
     ? {
-        tag: tag,
+        type: type,
         props: props || {},
         children: children
       }
-    : tag(props, children)
+    : type(props, children)
```
JorgeBucaran: Should add `props || {}` here, but maybe in a different PR 😭 😆

zaceno: Nah might as well fix such a tiny thing while we got it (or risk forgetting...) I'll do it

zaceno: But hang on. Is there an equivalent `children || []` somewhere else?

zaceno: Oh. yeah I see it

zaceno:
Ok I changed it. Good to merge now?

PR #41, opened 2017-11-06
Remove container default

### LINKED ISSUE #40: Remove container default
Just an *opinion* based on my personal learning experience with this library.

I'd like to suggest removing the default value for `container` in the `patch` function.

I found it quite surprising that there was any default at all - but since the default is `document.body`, your first experience, if you're following the example, will be that of your own document getting wiped out, which was pretty surprising to me.

Also, when looking at the example in the README for the first time, my first thought was, "if both of the arguments are virtual DOM nodes, how does it figure out which real DOM element to updated?"

Of course, you learn these things quickly if you start poking through the source-code, but most people will likely jump in without reading documentation, or copy/pasting the example - making this argument non-optional might remove any confusion about how a target element is established and/or what happened to your existing body-content.

Defaults don't always curb the learning curve - sometimes they increase it.

mindplay-dk:
Closes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ import { h, patch } from "picodom"
 let node
 
 function render(view, withState) {
-  patch(node, (node = view(withState)), document.body)
+  patch(document.body, node, (node = view(withState)))
 }
 
 function view(state) {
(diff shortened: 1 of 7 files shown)


JorgeBucaran:
@mindplay-dk LGTM, shall we merge?

Issue #18, opened 2017-07-22
Setting opacity to 0 


mrjjwright:
I set opacity to 0 as follows:

`h(
          "div",
          {
            class: "",
            style: { width: 15, height: 15, opacity: 0 },
            oninsert: function(el) {
              checkEl = el;
            }
          },
          "✔"
        )`

Picodom is not outputting an opacity css style prop on the element.  If I set it to 0.1 or higher it does.

jbucaran:
@mrjjwright Interesting! I think this might be a bug. I'll keep you posted on this.

mrjjwright:
Thanks, also it appears that strings like “0” work, which might be a clue.

yepbug:
I think [this line](https://github.com/picodom/picodom/blob/master/src/patch.js#L151) is why there is a bug (`value['opacity']` is 0, which is a false value) . And change this line to `element.style[i] = String(value[i]) || "";` can fix this.

jbucaran:
Thanks @yepbug. Yes, that should do it. 

@mrjjwright Just curious. Why were you using a number instead of a string?

acstll:
Here's some interesting discussion: https://github.com/w3c/webcomponents/issues/519

I think doing `String(value[i]) || ""` might not be the best answer, because other values like `false` or `null` would be stringified as well and you don't want `opacity = "null"`. Maybe something like testing against `value[i]` not being null or undefined?:

```js
element.style[i] = value[i] != null ? value[i] || ""
```
The most important thing would be to settle on the API first, should attributes accept only strings? how should other types be handled? And make it clear on the documentation.

jbucaran:
@acstll Why would you use a number instead of a string again?

acstll:
> @acstll Why would you use a number instead of a string again?

I can imagine getting the value out of a computation, for example, so you just use that. You do something like `{ opacity: result }`; it's just easier. One can always do `{ opacity: String(result) }` I guess.

I'm in favor of only accepting strings as attribute values, so this here wouldn't be a bug. And I'm just giving my opinion here I'm not even using picodom myself :-)

mindplay-dk:
IMO, disallowing `number` doesn't make sense - this is Javascript, and people *will* use numbers regardless of any specification or documentation. What would you do, throw an `Error`? That would likely be surprising to JS developers who naturally expect scripts to handle different types dynamically - it would just be pedantic.

[...]

PR #46, opened 2017-11-09
Handle numbers in `styles`

acstll:
This would solve #18 

You can write

```js
h("div", { style: { opacity: 0 } })
```

and get

```html
<div style="opacity: 0;"></div>
```

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -55,7 +55,7 @@ function setElementProp(element, name, value, oldValue) {
   if (name === "key") {
   } else if (name === "style") {
     for (var name in merge(oldValue, (value = value || {}))) {
-      element.style[name] = value[name] || ""
+      element.style[name] = value[name] != null ? value[name] : ""
     }
   } else {
     try {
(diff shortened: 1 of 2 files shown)


src/patch.js
```diff
@@ -55,7 +55,7 @@ function setElementProp(element, name, value, oldValue) {
   if (name === "key") {
   } else if (name === "style") {
     for (var name in merge(oldValue, (value = value || {}))) {
-      element.style[name] = value[name] || ""
+      element.style[name] = value[name] != null ? value[name] : ""
```
JorgeBucaran: @acstll Could we get away with?

```js
element.style[name]+=""
```

acstll: With `element.style[name]+=""` tests fail.

`!= null` checks against `null` and `undefined`, so `0` is valid. Anything else will be already coerced into a string by the browser.

If I'm correct, this change will keep things intact and only allow `0` to be valid and not being swallowed as falsy. Everything else stays the same.

JorgeBucaran: Sorry, I wasn't thinking well! 😅

JorgeBucaran:
Thanks @acstll! 👍

acstll:
You're welcome!

PR #50, opened 2017-11-12
Handle falsy attribute values correctly

### LINKED ISSUE #49: setElementProp removes falsy attribute values
In a recent `hyperapp` issue [Updating children props dynamically ignores "0" #419](https://github.com/hyperapp/hyperapp/issues/419) `setElementProp` behaviour was improved when dealing with falsy values.

Will `picodom` be getting the same logic update?

My specific use case is attempting `<option value="">Not applicable</a>` but `picodom` removes my value attribute and the browser subsequently considers the selected value to be `"Not applicable"`

Contrived example: https://codepen.io/maxholman/pen/vWmgPa

maxholman:
Closes #49 

Logic is identical to that of `hyperapp` at time of writing

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -63,10 +63,10 @@ function setElementProp(element, name, value, oldValue) {
     } catch (_) {}
 
     if (typeof value !== "function") {
-      if (value) {
-        element.setAttribute(name, value)
-      } else {
+      if (null == value || false === value) {
         element.removeAttribute(name)
+      } else {
+        element.setAttribute(name, value)
       }
     }
   }
--- a/src/patch.js
+++ b/src/patch.js
@@ -66,7 +66,7 @@ function setElementProp(element, name, value, oldValue) {
       if (null == value || false === value) {
         element.removeAttribute(name)
       } else {
-        element.setAttribute(name, value)
+        element.setAttribute(name, value === true ? "" : value)
       }
     }
   }
--- a/src/patch.js
+++ b/src/patch.js
@@ -59,7 +59,7 @@ function setElementProp(element, name, value, oldValue) {
     }
   } else {
     try {
-      element[name] = value
+      element[name] = null == value ? "" : value
     } catch (_) {}
 
     if (typeof value !== "function") {
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -601,7 +601,7 @@ testTrees("styles", [
   {
     node: h("div"),
     html: `<div style=""></div>`
-  },
+  }
 ])
 
 testTrees("update element data", [
@@ -703,5 +703,4 @@ testTrees("elements with falsy values", [
     }),
     html: `<input type="text">`
   }
-
-]);
+])
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -660,7 +660,7 @@ testTrees("update element with dynamic props", [
   }
 ])
 
-testTrees("elements with falsey values", [
+testTrees("elements with falsy values", [
   {
     node: h("input", {
       type: "text",
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -3,7 +3,7 @@ import { h, patch } from "../src"
 function testTrees(name, trees) {
   test(name, done => {
     let node
-    
+
     trees.forEach(tree => {
       patch(document.body, node, (node = tree.node))
       expect(document.body.innerHTML).toBe(tree.html.replace(/\s{2,}/g, ""))
@@ -659,3 +659,49 @@ testTrees("update element with dynamic props", [
     html: `<input type="text" value="foo">`
   }
 ])
+
+testTrees("elements with falsey values", [
+  {
+    node: h("input", {
+      type: "text",
+      value: ""
+    }),
+    html: `<input type="text" value="">`
+  },
+  {
+    node: h("input", {
+      type: "text",
+      value: "0"
+    }),
+    html: `<input type="text" value="0">`
+  },
+  {
+    node: h("input", {
+      type: "text",
+      value: 0
+    }),
+    html: `<input type="text" value="0">`
+  },
+  {
+    node: h("input", {
+      type: "text",
+      value: null
+    }),
+    html: `<input type="text">`
+  },
+  {
+    node: h("input", {
+      type: "text",
+      value: false
+    }),
+    html: `<input type="text">`
+  },
+  {
+    node: h("input", {
+      type: "text",
+      value: undefined
+    }),
+    html: `<input type="text">`
+  }
+
+]);


test/vdom.test.js
```diff
@@ -659,3 +659,49 @@ testTrees("update element with dynamic props", [
     html: `<input type="text" value="foo">`
   }
 ])
+
+testTrees("elements with falsy values", [
```
JorgeBucaran: @maxholman Is this your original test? 😄 

Make sure to run prettier on all the files 🙏

src/patch.js
```diff
@@ -63,10 +63,10 @@ function setElementProp(element, name, value, oldValue) {
     } catch (_) {}
```
JorgeBucaran: @maxholman You missed some code above the `catch` block.

maxholman: @JorgeBucaran Added :+1: Not 100% sure what it does but it didn't affect the test results or coverage %

test/vdom.test.js
maxholman: I wrote the test just for this PR yes. Has now been "prettified" :)

src/patch.js
JorgeBucaran: @maxholman That part makes sure that if value is null, we only assign `""` to the prop instead of `null` or `undefined`, which in the case of text inputs would be converted to a string.

mindplay-dk:
Should this also handle `true`?

I think all browsers will accept `setAttribute(name, true)`, but technically it's wrong - the type of the second argument is `string`, and according to the specs, the correct way to set a value-less attribute is with the empty string `""`.

JorgeBucaran:
@mindplay-dk To what line would you be referring to?

src/patch.js
mindplay-dk: @JorgeBucaran I was referring to this section, which now explicitly handles `false` (which is what this PR is about) but there is no explicit handling for `true`. I know it'll work, but per the [specs](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute), it's a `DOMString`, and "any non-string value specified is converted automatically into a string", which means you end up with an attribute-value of `"true"`, which is technically an invalid value for properties such as `readonly`, even if it does work.

JorgeBucaran: @mindplay-dk Are you referring to (1) or (2)?

1. https://github.com/picodom/picodom/pull/50/files#diff-2eda87b7e7907c5229573454410d120dR62
2. https://github.com/picodom/picodom/pull/50/files#diff-2eda87b7e7907c5229573454410d120dR66

mindplay-dk: The second - `changeElementProp()`

JorgeBucaran: @mindplay-dk Okay, now I fully understand what we are discussing.

So this instead?

```js
element.setAttribute(name, value === true ? "" : value)
```

I guess we could, but in this particular case the current behavior is harmless and effective. So, I don't want to handle `true` values.

mindplay-dk: Note that, if someone was to export or inspect the HTML, they would see `readonly="true"` rather than `readonly`. Also note that attempting to do it correctly, by trying to set the value to `""`, is currently prevented by the library. I think there ought to be at least some way to create correct markup? Also, there is the issue of consistency with other JSX frameworks.

You've said this wasn't about filesize, but about staying small and comprehensible. I hope we're not making tradeoffs to save a few bytes? Explicitly handling both `true` and `false` would probably raise fewer questions than not.

Also note, if you ever port this to TS, you will need at least an explicit type-assertion to suppress the error ;-)

JorgeBucaran: I am happy to create an issue and keep it open until we can be completely sure this is one of those "must have" or not. Is this one of those we should support or not is the question. If the answer is definitely yes, then let's do it 💪😄 , but the keep ind mind that:

- "Consistency with other JSX frameworks" is not in the scope of picodom, and
- "export or inspect the HTML", you can still do it and not be confused with the result, in fact it may be more readable. Why would you do that anyway? I'm just curious.

> Also note, if you ever port this to TS, you will need at least an explicit type-assertion to suppress the error.

Now, this is actually a good reason. If we do convert to TypeScript, then I'd rather avoid fooling the compiler just so I can have my cake and eat it.

I started to have mixed feelings about TypeScript after working a bit on Hyperapp's typings, and finding I was unable to appreciate it how it could improve my workflow. There's just so much "typing" and repeating stuff.

What do you think about flow?

Rayraegah: Could avoid it by using a type check before you remove the attribute (typeof null is object)

```javascript
try {
  element[name] = value === true ? "" : value;
} catch (err) {}

if (typeof value !== "function") {
  if (
    !value &&
    typeof value !== "string" &&
    typeof value !== "number"
  ) {
    element.removeAttribute(name);
  } else {
    element.setAttribute(name, value);
  }
}
```

mindplay-dk: In my opinion, Flow is basically Typescript, just with horrible syntax.

There will be just as much "typing" and repeating stuff - there will be a "compile" step to validate types and strip the annotations just the same. I know there are other differences between Flow and Typescript - the classic argument in favor of Flow was better null-handling, but that's since been solved with the introduction of strict mode in TS, which I highly recommend. 

One argument I keep hearing is "standards", but, meh - the Flow syntax may be embedded in standard JS comments, but it's still a proprietary syntax; I prefer Typescript which is much more readable (and faster to write) once you learn how to read it.

Also, TS has awesome IDE support.

I'd say, try it before you judge. Pico is the perfect size project to try it out and learn. I have ported several large JS libraries (including a parser library and a WYSIWYG editor) to TS, and in every case, I found the JS was littered with subtle bugs - introucing types is a really valuable way to vet and improve code correctness.

As to the amount of "typing" and repeating stuff, it's likely much less than you think - TS has inference for most things, and using VS code, you can hover over any function declaration, variable, etc. and see what type the compiler thinks it has, so you can avoid explicitly restating your assumptions. Once everything is typed, you can enjoy much safer and faster refactoring.

I don't know anyone who has learned it and regretted :-)

mindplay-dk: If you'd like to see what it would look like in TS, I'd be happy to give it a shot - porting would be a valuable exercise for me, as I'd get to learn every internal detail of the codebase, and I will likely do it sometime soon anyway, just as an exercise, since it's small enough :-)

JorgeBucaran: @mindplay-dk Thank you for the encouraging words! I am going to give it another shot. 🏄

mindplay-dk: Groovy 👍 

I'll hold off on porting it myself - it's possible that a type-hinted version (with more inline documentation perhaps?) will answer all my questions :-)

JorgeBucaran: @mindplay-dk Convinced me! 🙇 

@maxholman Can you change the offending line to:

```js
element.setAttribute(name, value === true ? "" : value)
```

maxholman: @JorgeBucaran line changed :+1: I can backport to hyperapp as well if it helps?

JorgeBucaran:
Merged! 🎉

PR #55, opened 2017-11-21
set class-attribute from Object or Array values

mindplay-dk:
Add support for initializing the class-attribute from a map where class-name maps to `true`/`false` values, or from an Array of class-names.

This is very common and very widely-used feature in most JSX-based libraries - it gets a bit cumbersome to handhold this for every view implementation, and this shouldn't add much to the minified footprint, I think?

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -57,6 +57,12 @@ function setElementProp(element, name, value, oldValue) {
     for (var name in merge(oldValue, (value = value || {}))) {
       element.style[name] = value[name] != null ? value[name] : ""
     }
+  } else if (name === "class" && typeof value === "object") {
+    let classNames = value.constructor == Object
+      ? Object.keys(value).filter(function(className) { return value[className] })
+      : value // assumes Array
+
+    element.setAttribute(name, classNames.join(" "))
   } else {
     try {
       element[name] = null == value ? "" : value
--- a/test/vdom.test.js
+++ b/test/vdom.test.js
@@ -604,6 +604,17 @@ testTrees("styles", [
   }
 ])
 
+testTrees("class names", [
+  {
+    node: h("div", { class: { foo: true, bar: true, qux: false } }),
+    html: `<div class="foo bar"></div>`
+  },
+  {
+    node: h("div", { class: ["foo", "bar"] }),
+    html: `<div class="foo bar"></div>`
+  }
+])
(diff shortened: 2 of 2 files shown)


JorgeBucaran:
@mindplay-dk At first glance this seems like the kind of thing you get by using https://github.com/JorgeBucaran/classcat or JedWatson/classNames yes?

mindplay-dk:
I guess, though both of these do more than just join class-names?

Makes me wonder if we ought to put property-setters in a map and expose it? If we could plug in custom property-setters, people could choose from these packages, for example. (I'd also like to have support for e.g. `data` and the `on`-attribute that most JSX frameworks support, at some point - I find these are all the "bare bones" no-brainer features that everyone needs. "batteries included.")

Though in the case of the class-attribute, IMO, it's common enough to include by default - practically everyone would need to install and plug in another package for this.

JorgeBucaran:
@mindplay-dk I think this is not acceptable for picodom. I don't expect picodom to be used for building frontend apps; I didn't build it for that. I expect picodom to be used as a building block to create lightweight JS frameworks, very much like Hyperapp.

mindplay-dk:
> I expect picodom to be used as a building block to create lightweight JS frameworks

I get it! "wrong layer", right? Things like class-name transformations and other property/attribute magic, I can absolutely implement in a higher-level layer that makes these transformations *before* passing them to picodom. Gotcha :-)

What this doesn't address though, is the difference between properties and attributes - the need to know or decide which you're updating.

DOM elements ultimately are a duality: the objects have a set of properties *and* a set of attributes - unfortunately, JSX only has one syntax for attributes, so we need to address this somehow.

Currently, the VDOM model doesn't allow you to distinguish between properties and attributes, at all. That's a problem - if I'm building a micro framework on top of pico, I don't have any control over properties and attributes individually.

We should open a new issue to discuss this further maybe?

mindplay-dk:
I've opened a separate issue, so I'm closing this one.

JorgeBucaran:
@mindplay-dk Yeah, let's open a new issue to discuss this very interesting topic. I have a few opinions on the subject! 🙏

JorgeBucaran:
@mindplay-dk You already opened it! https://github.com/picodom/picodom/issues/56 :)

PR #59, opened 2017-11-24
Minor improvements to picodom.d.ts

### LINKED ISSUE #47: Typescript: add default type-argument to VNode
Having to explicitly type-hint node arguments as `Node<{}>` gets tedious doesn't read well.

Can we make it the default?

```ts
  type: string
  props?: Props
  children: Array<VNode<{}> | string>
}
```

This way you can type-hint as just `Node`.

We can then clean up all the other `VNode<{}>` type-hints in the declarations file as well.

mindplay-dk:
Use default type-argument of `{}` for VNode type, remove redundant type-arguments.

Also, make `patch` accept `Element` rather than `HTMLElement` - an `SVGElement` is not an `HTMLElement`, so this would get in the way of patching SVG element; it's also inconsistent with the return-type.

Closes #47

### DIFF
--- a/picodom.d.ts
+++ b/picodom.d.ts
@@ -1,31 +1,31 @@
 export as namespace picodom
 
-export interface VNode<Props> {
+export interface VNode<Props = {}> {
   type: string
   props?: Props
-  children: Array<VNode<{}> | string>
+  children: Array<VNode | string>
 }
 
 export interface Component<Props = {}> {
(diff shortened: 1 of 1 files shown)


JorgeBucaran:
Hey guys, @andrewiggins @pspeter3 @swizz, would have a look? 👋

JorgeBucaran:
Shall we merge? :)

PR #70, opened 2017-12-19
Property precedence

### LINKED ISSUE #56: Indistinguishable properties and attributes
Currently, the `patch()` function treats every JSX attribute simultaneously as a property *and* an attribute - it initializes both, which means you have no ability to update them individually.

Blindly treating every JSX attribute as both a property and an attribute means that most updates are unnecessarily being done twice - one of them generating either a meaningless property or a meaningless DOM attribute. The DOM accumulates a lot of garbage this way.

For example, `<div innerHTML={content}>` will inject literal HTML via the `innerHTML` property, but will also create a useless `innerHTML` *attribute* - or in some cases, such as `<input value={value}>`, will initialize the value twice, first using the `value` property, then again using `setAttribute('value', ...)`.

I don't have a lot of references, but `snabbdom` for one [separates](https://github.com/snabbdom/snabbdom/blob/master/src/vnode.ts#L23) properties from attributes - which seems sort of inevitable, at the VDOM level, if we want to solve this problem?

mindplay-dk:
Fixes #56

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -58,20 +58,18 @@ function setElementProp(element, name, value, oldValue) {
       element.style[name] = value[name] != null ? value[name] : ""
     }
   } else {
+    let empty = null == value || false === value
+
     if (name in element) {
       try {
         element[name] = value == null ? "" : value
       } catch (_) { }
-
-      if (null == value || false === value) {
-        element.removeAttribute(name)
-      }
-    } else if (typeof value !== "function") {
-      if (null == value || false === value) {
-        element.removeAttribute(name)
-      } else {
-        element.setAttribute(name, value === true ? "" : value)
-      }
+    } else if (!empty && typeof value !== "function") {
+      element.setAttribute(name, value === true ? "" : value)
+    }
+    
+    if (empty) {
+      element.removeAttribute(name)
     }
   }
 }
--- a/src/patch.js
+++ b/src/patch.js
@@ -1,12 +1,7 @@
 var callbacks = []
 
 export function patch(parent, oldNode, newNode) {
-  var element = patchElement(
-    parent,
-    parent.children[0],
-    oldNode,
-    newNode
-  )
+  var element = patchElement(parent, parent.children[0], oldNode, newNode)
 
   for (var cb; (cb = callbacks.pop()); cb()) {}
 
@@ -63,11 +58,11 @@ function setElementProp(element, name, value, oldValue) {
     if (name in element) {
       try {
         element[name] = value == null ? "" : value
-      } catch (_) { }
+      } catch (_) {}
     } else if (!empty && typeof value !== "function") {
       element.setAttribute(name, value === true ? "" : value)
     }
-    
+
     if (empty) {
       element.removeAttribute(name)
     }
--- a/src/patch.js
+++ b/src/patch.js
@@ -49,8 +49,8 @@ function createElement(node, isSVG) {
 function setElementProp(element, name, value, oldValue) {
   if (name === "key") {
   } else if (name === "style") {
-    for (var name in merge(oldValue, (value = value || {}))) {
-      element.style[name] = value[name] != null ? value[name] : ""
+    for (var i in merge(oldValue, (value = value || {}))) {
+      element.style[i] = value[i] != null ? value[i] : ""
     }
   } else {
     let empty = null == value || false === value
--- a/src/patch.js
+++ b/src/patch.js
@@ -59,14 +59,14 @@ function setElementProp(element, name, value, oldValue) {
     }
   } else {
     try {
-      element[name] = value
+      element[name] = null == value ? "" : value
     } catch (_) {}
 
     if (typeof value !== "function") {
-      if (value) {
-        element.setAttribute(name, value)
-      } else {
+      if (null == value || false === value) {
         element.removeAttribute(name)
+      } else {
+        element.setAttribute(name, value === true ? "" : value)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@mindplay-dk Now that you understand the issue better (I presume) can you summarize why this change is necessary and what benefit do we get?

mindplay-dk:
The main benefit is we no longer pollute the DOM with meaningless attributes, nor do we waste resources doing so - for example, before, if someone were to set the `className` property, you'd also end up with a meaningless `className` attribute in the DOM.

See also description of #56 for details.

mindplay-dk:
Note that `uglify` currently fails, so this is probably not ready for merge.

```
> uglifyjs dist/picodom.js -o dist/picodom.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --in-source-map dist/picodom.js.map --source-map dist/picodom.js.map


Parse error at picodom.js:95,8

    let empty = null == value || false === value;

        ^

SyntaxError: Unexpected token: name (empty)

    at JS_Parse_Error.get (eval at <anonymous> (/home/travis/build/picodom/picodom/node_modules/uglify-js/tools/node.js:27:1), <anonymous>:86:23)

    at /home/travis/build/picodom/picodom/node_modules/uglify-js/bin/uglifyjs:384:39

    at time_it (/home/travis/build/picodom/picodom/node_modules/uglify-js/bin/uglifyjs:620:15)

    at /home/travis/build/picodom/picodom/node_modules/uglify-js/bin/uglifyjs:345:9

    at tryToString (fs.js:449:3)

    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:436:12)
```

I've never used `uglify`, but perhaps this is due to the fact I automatically use ES2016 syntax by now? Looks like the tests are targeting an older version of Node?

src/patch.js
```diff
@@ -58,16 +58,18 @@ function setElementProp(element, name, value, oldValue) {
...
-      element[name] = null == value ? "" : value
-    } catch (_) {}
-
-    if (typeof value !== "function") {
-      if (null == value || false === value) {
-        element.removeAttribute(name)
-      } else {
-        element.setAttribute(name, value === true ? "" : value)
-      }
+    let empty = null == value || false === value
```
JorgeBucaran: `var`

JorgeBucaran:
@mindplay-dk Would the following work as well?

```js
function setElementProp(element, name, value, oldValue) {
  if (name === "key") {
  } else if (name === "style") {
    for (var name in merge(oldValue, (value = value || {}))) {
      element.style[name] = value[name] != null ? value[name] : ""
    }
  } else {
    var empty = null == value || false === value

    if (name in element) {
      try {
        element[name] = value == null ? "" : value
      } catch (_) {}
    } else if (typeof value !== "function") {
      if (null == value || false === value) {
        element.removeAttribute(name)
      } else {
        element.setAttribute(name, value === true ? "" : value)
      }
    }
  }
}
```

mindplay-dk:
What's the difference? Just `var` instead of `let`?

That'll work of course. I'm in the habit of using var/let/const by now, it's hard not to 🙄

JorgeBucaran:
@mindplay-dk No, I mean the code I posted above.

mindplay-dk:
Looks functionally equivalent to me, but maybe change it and run the test?

JorgeBucaran:
@mindplay-dk Is there any way we could simplify or compact this code a bit more? It's adding 20 bytes. It seems that setting some properties, like `id`, will cause the attribute to be set automatically, so that's why we need:

```js
if (empty) {
  element.removeAttribute(name)
}
```

...near the end, regardless of previous conditions.

mindplay-dk:
I'm honestly not interested in the file-size - I'm interested in simple, idiomatic, readable code.

If you can do both, that's fine - but "pre-minifying" the code really doesn't help with comprehension.

Isn't there a tool that can aggressively rewrite if/else clauses into compounded ternary expressions and so on? It's really not helpful to anybody - especially given how young this project still is...

Just my two cents.

JorgeBucaran:
@mindplay-dk Not that kind of easy minification, but actualy simplification by rewriting the logic altogether like I did. That didn't really work though, so this looks good. Could you run npm format (or make sure to run prettier on all the files you changed) so I can merge this?

Also, I left you a comment.

src/patch.js
```diff
@@ -58,16 +58,18 @@ function setElementProp(element, name, value, oldValue) {
       element.style[name] = value[name] != null ? value[name] : ""
```
JorgeBucaran: Not part of this PR, but could you change this to `element[name][i]` and also the loop variable from `name` to `i`?

mindplay-dk: Oh, I see, it's overwriting the name variable! Yeah, I'll fix that.

src/patch.js
```diff
@@ -54,20 +49,22 @@ function createElement(node, isSVG) {
 function setElementProp(element, name, value, oldValue) {
   if (name === "key") {
   } else if (name === "style") {
-    for (var name in merge(oldValue, (value = value || {}))) {
-      element.style[name] = value[name] != null ? value[name] : ""
+    for (var i in merge(oldValue, (value = value || {}))) {
+      element.style[i] = value[i] != null ? value[i] : ""
```
JorgeBucaran: @mindplay-dk 
```js
element[name][i]
```

mindplay-dk: Why would you want to do that dynamically? We know it'll only work with `style`, and it's an isolated use-case.

JorgeBucaran: It saves a few bytes.

JorgeBucaran:
Thanks @mindplay-dk! 🎉 😄

src/patch.js
mindplay-dk: 2 to be exact 😆

JorgeBucaran:
@mindplay-dk I ported this change to Hyperapp and it broke SVG! 😆 #509

PR #72, opened 2017-12-22
Add ondestroy & refactor functions in show order.

JorgeBucaran:
- Add new ondestroy lifecycle event called both for nodes that are directly removed or removed as a result of a parent (any ancestor) node being removed.

- Breaking: Change the type of onremove from returning a function that receives `done`, to a function that takes both the element and `done`. onremove(element, done).

Sorry! I committed a bunch of unrelated stuff into this PR, but if I don't move fast, then picodom will never reach 2.0.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -12,8 +12,9 @@ export function h(type, props) {
       for (i = node.length; i--; ) {
         stack.push(node[i])
       }
-    } else if (node != null && node !== true && node !== false) {
-      children.push(typeof node === "number" ? (node = node + "") : node)
+    } else if (null == node || true === node || false === node) {
+    } else {
+      children.push(typeof node === "number" ? node + "" : node)
     }
   }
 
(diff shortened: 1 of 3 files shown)


JorgeBucaran:
@mindplay-dk I believe I sent you an invitation to the org, did you receive it?

mindplay-dk:
> I believe I sent you an invitation to the org, did you receive it?

Hmm, nope?

JorgeBucaran:
@mindplay-dk No email notification either? Check out: https://github.com/picodom, you should see the invitation under the pine tree 🎄

mindplay-dk:
under the pine tree 😂

mindplay-dk:
Looks good, I guess - it's a bit hard to see what's changed, since everything also changed places... separate commits when moving things around next time pls, so we can read the diff?

I can take another pass at `ondestroy` after this is merged - it'll conflict for sure because of the reordering. Or you could review my changes (which will diff) and integrate them? I think my destroy notifications code is smaller? And the test a bit stricter.

Unfortunately this is the end of my day, and it's likely I won't be back to work on this before friday two weeks from now.

PR #73, opened 2017-12-22
add recursive `ondestroy` notification

### LINKED ISSUE #61: Missing `ondestroy` event for deep removals
It appears the `onremove` event isn't triggered for children, when the parent is removed - I believe the following test demonstrates the problem:

https://github.com/mindplay-dk/picodom/commit/f108584890bdc936b44e8db5c071384bae3f3476

Looks like a bug? I'm unsure how to address it.

mindplay-dk:
Fixes #61 

This addresses the need to integrate and clean up after third-party plain JS (or jQuery, etc.) components.

It's a temporary fix, in the sense that we know about the problems with `onremove` as it relates to `ondestroy` and the element order issue.

But I believe this gets me to the point where I can actually start to build a real project with picodom, seeing as how animation (via `onremove`) isn't really essential to shipping a functioning app.

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -86,7 +86,9 @@ function updateElement(element, oldProps, props) {
   }
 }
 
-function removeElement(parent, element, props) {
+function removeElement(parent, element, node) {
+  var props = node.props
+
   if (
     props &&
     props.onremove &&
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
@mindplay-dk This is great work, but I'd like to proceed with my PR here #72! If you could add your review comments that would be great! 🙏

PR #75, opened 2017-12-29
use `bundlesize` to validate gzipped footprint

mindplay-dk:
Looks like we're over the magical 1KB limit.

@JorgeBucaran tbh, I'm submitting this PR a bit poignantly - I hope this doesn't prompt you to run off and start mangling or crippling the code to get under the 1KB limit ;-)

But this PR will currently break the build, as we're already over the limit.

It might be useful as a means of tracking size before/after changing (or refactoring for size) to see what really works or doesn't - due to the nature of gz, the byte-size isn't always directly dependent on the byte-sizes pre-compression before/after making a certain change, so this could help with that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "test": "jest --coverage --no-cache && tsc -p test/typings/tsconfig.json",
-    "build": "npm run bundle && npm run minify",
+    "build": "npm run bundle && npm run minify && bundlesize",
     "bundle": "rollup -i src/index.js -o dist/picodom.js -f umd -mn picodom",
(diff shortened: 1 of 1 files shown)


leeoniya:
fyi, node can do gzip natively via zlib, so you don't need an extra dependency.

here's how i do it:

https://github.com/leeoniya/domvm/blob/3.x-dev/build.js#L7
https://github.com/leeoniya/domvm/blob/3.x-dev/build.js#L189-L193

mindplay-dk:
@leeoniya thanks! there are likely several other things we could simplify/improve with a custom build script, but we don't currently have one, and I didn't want to introduce one with this PR.

JorgeBucaran:
@mindplay-dk Sorry to have to dismiss another of your contributions, I really like what you are doing, but I've decided against bundlesize already. Please see https://github.com/hyperapp/hyperapp/commit/4e45a7474b807ff8b333bd0fbfc6eaaa35b45d57

mindplay-dk:
No worries, this was 5 minutes of work :-)

PR #76, opened 2018-01-02
Updated test-case

mindplay-dk:
Updated my test-case according to your depth-first implementation.

I think it's good to have a test-case that covers both what should happen and what shouldn't - also, this covers multiple levels and the fact that it's depth-first.

(I'd suggest a squash-merge - the commit history is just noise at this point.)


JorgeBucaran:
Thanks, I'll merge here right away, add some changes and publish later! 👍

PR #78, opened 2018-01-03
improve and add missing docs

mindplay-dk:
Added the missing `ondestroy` event, updated the docs for the `onremove` event, clarified use of events with use-case examples, moved the `patch` function documentation up with the rest of the API description in the "Usage" section, documented the `key` attribute and when/how to use it, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,9 +93,27 @@ const element = patch(
 
 The `patch` function returns the patched child element.
 
-## Reserved Attributes
+## Supported Attributes
 
-Picodom supports element-level life-cycle events and keyed updates, and reserves the following JSX attributes:
+This section describes the particular handling of certain HTML/SVG attributes, and certain special attributes reserved by Picodom.
+
+### Standard HTML and SVG Attributes
+
+All standard HTML and SVG attributes are supported, and the following standard attributes are handled specifically:
+
+#### `class`
+
+Both the `className`-property and `class`-attribute are supported.
+
+#### `style`
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,43 +81,69 @@ function AwesomeTitle(text) {
...
+)
+```
+
+The `patch` function returns the patched child element.
+
+## Lifecycle Events
+
+Picodom supports element-level life-cycle events and keyed updates, and reserves the following JSX attributes:
+
+#### `key`
```
JorgeBucaran: @mindplay-dk `key` is not a lifecycle event though. Can you put this somewhere else?

mindplay-dk: But it's part of the life-cycle control, right?

It has no use other than to control the life-cycle of the element, I think?

So maybe the title of that section should be "Life-cycle Control", not "Life-cycle Events"?

Or maybe just "Reserved Attributes"?

JorgeBucaran: Yeah, reserved attributes sounds 👍

mindplay-dk:
@JorgeBucaran I'd like to document support for `class` and special handling of `style` too, so that all of attribute handling is fully documented - do you think I should do that under the same heading as the other attributes, or separately?

Maybe change the heading to something more general like "Supported Attributes" or "Special Attributes"?

JorgeBucaran:
@mindplay-dk Sounds good. "Supported Attributes" is fine 👍

mindplay-dk:
Good to go?

JorgeBucaran:
Good good! 🎉

PR #83, opened 2018-01-05
never set SVG attributes as props

### LINKED ISSUE #80: Debug SVG attributes
Basically just apply these changes from Hyperapp:

https://github.com/hyperapp/hyperapp/commit/4f147c2ec00fd2a69b2298981a81372c603211ca#commitcomment-26628985

mindplay-dk:
Applies [this change](https://github.com/hyperapp/hyperapp/commit/4f147c2ec00fd2a69b2298981a81372c603211ca?diff=unified#diff-1fdf421c05c1140f6d71444ea2b27638) from Hyperapp.

Closes #80

Note that there's no impact on tests, as we don't have tests for SVG support.

There is a potential minor problem with this approach - it may be acceptable, but seems a bit wrong, we're now treating *all* SVG properties as `string` types, which isn't quite right.

For example, the `width` and `height` properties might be computed as a `number`, which means [this test](https://github.com/picodom/picodom/blob/master/src/patch.js#L76) will fail - which effectively means that all numeric SVG properties initialized with numbers will be converted to strings and updated with every call to `patch`, because the type has "changed", even if the value has not; since the comparison is `!==`.

If you're okay with that, this PR is probably ok.


JorgeBucaran:
> @mindplay-dk ...all numeric SVG properties initialized with numbers will be converted to strings...

Can you show me where? I still don't see it.

Issue #68, opened 2017-12-13
Unexpected reordering after keyed removal


mindplay-dk:
Per this example using current unstable picodom `@2.0.0`

https://jsfiddle.net/4eunk2um/

Click to remove elements from the list, and note how the removed element always ends up at the bottom of the range - because it doesn't get destroyed until 2 seconds later.

Keyed updates presumably exist, for one, so you can do this sort of thing?

JorgeBucaran:
@mindplay-dk It seems this demos the problem you were having or am I looking at something else?

JorgeBucaran:
Nevermind, I see the elements are correctly removed, but I admit there's something funny going on.

zaceno:
@mindplay-dk @JorgeBucaran this is just how the algorithm works. We can't leave the to-be-removed-elements in their original place, because that messes up the indexes mapping vdom to DOM. So the elements go to the bottom of the list, outside the range of the vdom.

I don't think I'd consider it a bug so much as the only reasonable way to handle delayed element removal. More like a "known limitation" 

Incidentally, this behavior is precisely the reason I have to do position tracking in hyperapp-transitions.

mindplay-dk:
> We can't leave the to-be-removed-elements in their original place, because that messes up the indexes mapping vdom to DOM.

I think maybe we just discovered why some libs (petit-dom, snabbdom, etc.) implement diffing and patching in two separate passes.

To me, this is by no means a "limitation" - it makes `onremove` not useful for the one thing I'd want to use it for: transitions.

JorgeBucaran:
@mindplay-dk What would you suggest?

mindplay-dk:
One option is promises, like I suggested earlier - so that you can actually know when an update has finished. Not completely convinced that would even solve it necessarily though, since another update could be triggered indirectly by another DOM event (click, input, etc.) before the update is complete - you'd have to control the updates at a higher level to address that, e.g. something like snabbdom's `patch(container, vnode)` which "cheats" by injecting historical state into the `vnode`, which seems really dirty; it wouldn't be my preferred approach... there is surely another way to factor the same functionality though, without relying on side-effects, but anyhow...

I'm not sure there are any *simple* answers here.

[...]

PR #87, opened 2018-01-07
documents unexpected reordering per #68

mindplay-dk:
This isolated test-case documents #68, and could be merged as a regression-test when the issue gets fixed.


jorgebucaran:
I think this might be fixed now.

PR #88, opened 2018-01-07
add test for file size constraint

mindplay-dk:
Per the title, this adds a test for the 1KB file-size constraint.

It won't fail the test on overage, it'll only warn you on the console:

```
 PASS  test/filesize.test.js
  ● Console

    console.warn test/filesize.test.js:9
      File size: 1167 (143 over 1024 limit)
```

My gut feeling is there probably isn't much we can do, as everything is pretty compact already, and as said, I'd hate to the see code become more garbled (or features to get cut!) in favor of byte-size savings.

Maybe this will change if @JorgeBucaran gets around to rewriting a simpler patch function?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "babel-preset-env": "^1.6.0",
+    "gzip-size": "^4.1.0",
     "jest": "^21.2.1",
     "prettier": "~1.7.4",
     "rollup": "^0.50.0",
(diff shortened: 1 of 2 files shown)


JorgeBucaran:
Haha there is no 1024 limit. I'll review this later, thanks!

mindplay-dk:
Well, you advertise it everywhere as "the 1KB VDOM library", so... 😉 

Yeah, I'd be happy to officially drop (or raise) that constraint, so we can focus on the library itself rather than wasting time on marginal byte-size savings...

JorgeBucaran:
I don't think we need a constraint. The library is 1 KB because I am only using natural numbers. The library will never be 2 KB, so as long as we are under say 1500 bytes I am okay. Now, I don't think there is so much more I want to add to this.

Issue #82, opened 2018-01-05
`oncreate` vs `onupdate`


mindplay-dk:
I noticed that `onupdate` doesn't fire during the first update, is this intentional?

I think, in most systems that offer both a create and update notification, the initial creation triggers both a create and an update notification, because there isn't typically much use for an update notification that doesn't fire the first time.

Typically the notifications you need are for creation, for destruction, or for *all* updates - rarely does anyone need a notification for "all updates except the first one".

Thoughts?

JorgeBucaran:
> I think, in most systems that offer both a create and update notification, the initial creation triggers both a create and an update notification, because there isn't typically much use for an update notification that doesn't fire the first time.

I've noticed sometimes you need to use both oncreate/onupdate for the same thing and sometimes don't. When you have to, it's inconvenient writing a function with some logic, and calling it in two difference places. 

> in most systems that offer both a create and update notification

Do you have a concrete example? 

Also cc'ing this to @zaceno.

mindplay-dk:
> I've noticed sometimes you need to use both `oncreate`/`onupdate` for the same thing and sometimes don't. 

That's what I mean. I think this change would take care of that?

I guess it's a matter of what you think "on update" means - does it mean "updating the element", or "updating the attributes of the element"?

I think the latter makes more sense - because we don't really "update the element"; there's nothing you *can* update, other than it's properties and attributes, so we're really "updating the attributes of the element", which happens both after creating and updating the elements.

Can you think of a use-case where you need to do something when the attributes get udpated *except* the first time? I can't. Almost every use-case I can think is either creation only, or *every* update.

JorgeBucaran:
> Can you think of a use-case where you need to do something when the attributes get udpated except the first time?

I can't either. 🤔

zaceno:
@JorgeBucaran @mindplay-dk 

I haven't had situations where I need to do something on update except the first time.

What I have found, are usually I have to do the update-thing + something extra in oncreate. 

I suppose it could work as well to have onupdate run also the first time, but there would have to be a predictable order, like: oncreate always runs before onupdate. 

Except that will become tricky if, when the element is first created I want to do:

1) some stuff
2) the same stuff I want done every onupdate
3) some more stuff

In that case I'll have to move 2-3 into the onupdate, and use an If statement to detect if this is the first run of onupdate and only do 3 in that case.

This might not be very common though.

[...]

PR #90, opened 2018-01-09
Corrected description of remove

mindplay-dk:
Per closing of #82 this needs to accurately describe how it currently works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Fired after the element is created and attached to the DOM.
 
 #### `onupdate`
 
-Fired after the element attributes are updated. This event will fire even if the attributes have not changed.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,7 +133,7 @@ Fired after the element is created and attached to the DOM.
 
 #### `onupdate`
 
-Fired after the element attributes are updated. This event will fire even if the attributes have not changed.
+Fired after the existing element is updated. This event does *not* fire with `oncreate`, but *will* fire during element updates, regardless of whether any attributes have actually changed.
```
JorgeBucaran: > Fired after the existing element is updated. This event will fire even if the attributes have not changed.

mindplay-dk: I don't understand? You're just quoting the original text?

The point was to clarify the fact that `oncreate` and `onupdate` are mutually exclusive.

Issue #89, opened 2018-01-08
Remove `parent` as 1st mandatory argument for patch


JSteunou:
Having `parent` as first argument works very well when the all app is designed and wrote with picodom but not that much when used in a mixed libs app. (yes it happens in real world)

You cannot then just *render / create* a component to get its main node and then append it manually to a parent node of your choice at the time you choose.

To get around this you have to create a fake parent with `document.createElement` but that add a useless layer.

I'm also not a big fan of that kind of dependency injection and lake of separation which introduce a dependence between the child and its parent, I do prefer the other way, parent knowing its children, but I think this is just a matter of taste.

estrattonbailey:
Actually hitting the same road block just this morning 😛

I'm working on a localized stateful component using picodom and the assumption that `parent.children[0]` is the root of the app from [`patchElement`](https://github.com/picodom/picodom/blob/master/src/patch.js#L113) does make it tricky. My current solution essentially runs a normal render on the first pass, and then subsequent state updates are "rendered" like this:

```javascript
  // this._self - previous vdom node
  // this.render() - abstracted render fn
  // this.ref - outer DOM node of the component
  const temp = document.createElement('div')
  temp.appendChild(this.ref.cloneNode(true))
  patch(temp, this._self, (this._self = this.render(this.props, this.state)))
  const next = temp.childNodes[0]
  this.ref.parentNode.replaceChild(next, this.ref)
  this.ref = next
```

Not ideal, though I haven't run any benchmarks to really understand the performance impact here. It may be negligible in the grand scheme of things.

Would it be possible to expose a subset of the `patchElement` function that is reused internally *and* could be used externally to diff subtrees?

JorgeBucaran:
Thanks, @JSteunou and @estrattonbailey for bringing this up. Do you have any proposed alternatives? 🤔

estrattonbailey:
Looking into it, seems like a fun challenge. I've pulled down this repo and linked it to my current project but my experience writing VDOM reconcilers is... well it's zero, lol. Will report back if I have a suggestion.

JSteunou:
I really like `patch(node, newVNode)` returning `node`, seems the most simple from a user perspective, and the most *I dont care about vdom, just patch my node*

JorgeBucaran:
Can you be more exhaustive? What are the types of those arguments you wrote there? Where is the container?

JSteunou:
[...]

PR #91, opened 2018-01-09
API update - mount() for mounting, patch() for patching

estrattonbailey:
As mentioned in #89, the current `patch()` API essentially does double duty: mounting to the DOM at a single node, and updating the tree as the application updates.

This PR aims to introduce a separate `mount()` function and updates the `patch()` function's API to support patching at any level of the tree.

```javascript
/** @jsx h */
import { h, mount, patch } from 'picodom'

const App = props => <h1>{props.title}</h1

let root = <App title='Hello world!' />

mount(document.body, root)

// after state updates
patch(root, (root = <App title='New title' />))
```

As you can see, it adds a few lines, but perhaps there is a more compact way 🤔 

As far as introducing a new method to the API, I think it's a good idea. "Mounting" and "patching" are different concerns, and a single API for both would be confusing IMO.

## TODO
If you like this direction, I'm happy to make any updates necessary to get it integrated 🎉 

The tests are failing currently, I could probably use some direction there: `parentNode` is undefined. Throughout the suite `patch` would need to be updated to `mount` and the params slightly rearranged.

Thanks!

### DIFF
--- a/src/patch.js
+++ b/src/patch.js
@@ -1,11 +1,19 @@
 var callbacks = []
 
-export function patch(parent, oldNode, newNode) {
-  var element = patchElement(parent, parent.children[0], oldNode, newNode)
+export function mount(parent, node) {
+  var element = parent.appendChild(createElement(node))
 
   for (var cb; (cb = callbacks.pop()); cb()) {}
 
   return element
+}
+
+export function patch(oldNode, newNode) {
+  newNode._ref = patchElement(oldNode._ref.parentNode, oldNode._ref, oldNode, newNode)
+
+  for (var cb; (cb = callbacks.pop()); cb()) {}
+
+  return newNode._ref
 }
 
 function copy(target, source) {
@@ -65,6 +73,8 @@ function createElement(node, isSVG) {
     for (var i in node.props) {
       setElementProp(element, i, node.props[i])
     }
+
+    node._ref = element
   }
   return element
 }
--- a/src/index.js
+++ b/src/index.js
@@ -1,2 +1,2 @@
 export { h } from "./h"
-export { patch } from "./patch"
+export { patch, mount } from "./patch"


estrattonbailey:
# [Demo](https://github.com/estrattonbailey/picodom-demo)
Installs the library from my repo and mounts a tiny app and stateful component.

src/patch.js
```diff
@@ -1,13 +1,21 @@
...
+export function mount(parent, node) {
+  var element = parent.appendChild(createElement(node))
 
   for (var cb; (cb = callbacks.pop()); cb()) {}
 
   return element
 }
 
+export function patch(oldNode, newNode) {
+  newNode._ref = patchElement(oldNode._ref.parentNode, oldNode._ref, oldNode, newNode)
```
JorgeBucaran: @estrattonbailey Can we avoid doing this? I don't want to save stuff like this on the node. 🤔

estrattonbailey: Ah, I meant to mention that above. The main issue I saw with patching subtrees was that there was no reference to the real DOM node for any given tree of virtual nodes, so `patchElement` didn't have any root node to mutate childnodes from.

Do you think there's another way to get a DOM reference for an arbitrary virtual node? In terms of memory consumption, maybe a `getter` would help. Or maybe you see an entirely different way of doing this?

mindplay-dk:
Here's the approach I've been using:

```js
function mount(el, update) {
  let old
  
  return function() {
    patch(el, old, old = update())
  }
}
```

The function creates a closure for a target element and an update function, which is any function that returns a VNode. Here's an example with a stateless view:

https://jsfiddle.net/mindplay/tuq42u7t/

Because `update` is a function, this works equally well for views that require a `model` argument, for a functional component that requires `props` and `children`, or pretty much anything.

jorgebucaran:
@estrattonbailey Thanks, but I went with this instead:

```jsx
import { h, patch } from "picodom"

document.body.appendChild(
  patch( h("h1", {}, "Hello") )
)
```

As you can see, the patch function now returns an element, when none is given, allowing you to append, if that's the behavior you prefer. Otherwise, if you pass an element, patchdom will attempt to patch **that** element, instead of using it as a container like in previous versions.

estrattonbailey:
@jorgebucaran beautiful 😍

I should have renamed this PR, lol. My main goal here was was enable patching a VDOM node **at** a DOM node target. My tweak to `patch` made a new `mount` interface useful. It looks like your new version does exactly this, using similar assumptions to mine.

Can't wait to dive into this, 99% positive this is exactly what I needed. Thanks for your work!

Issue #94, opened 2018-02-13
2.0 API


JorgeBucaran:
Picodom@2.0.0 is coming to you very soon. The signature of <samp>patch</samp> needs to change, and I need your feedback to get it right.

Closes
- https://github.com/picodom/picodom/issues/80: I don't understand this issue.
- https://github.com/picodom/picodom/issues/77: I will add new tests.
- https://github.com/picodom/picodom/issues/92: Either one of below proposals will enable you to target a DOM element to be replaced instead of always pre-pending to a container.
- https://github.com/picodom/picodom/issues/89: The "Double Espresso Steamed Soy Milk Foam" proposal below fixes this issue, but I'll likely proceed with "Bitter Tea" and close that issue as wontfix. There is no way to please everyone.
- https://github.com/picodom/picodom/issues/65: I'll give a shot at adding types for VDOM events and close there.

### Bitter Tea

Bring back the old pre-1.0 signature. This API is more flexible and easier to implement at the expense of a slightly worse developer experience (is it?). The justification is that Picodom is intended to be used to create tiny Hyperapp-like view frameworks instead of directly consumed. Still, this is IMO much better than similar popular alternatives.

```js
import { h, patch } from "picodom"

// First and every other patch.
let element = patch(
  container,
  element,
  oldNode,
  newNode
)
```

### Double Espresso Steamed Soy Milk Foam

This proposal arguably improves the surface API providing at the expense of "slightly" less flexibility. This also has out of the box SSR DOM rehydration. This API makes consuming picodom directly more fun, but a bit awkward to integrate into your own library because you need to be aware there would be essentially two ways to call <samp>patch</samp>, the first time you call it, and every other time.

SSR recycling is enabled by providing a <samp>recycleElement</samp> to patch that works the same way as https://github.com/hyperapp/hyperapp/pull/590.

```js
import { h, patch } from "picodom"

// First patch.
let element = patch(vnode, containerElement)

// or first patch with server side rendered rehydration

let element = patch(vnode, containerElement, recycleElement)

// Every other patch.
patch(vnode, element)
```

/cc  @mindplay-dk @JSteunou  @estrattonbailey @pspeter3 @maxholman @dancespiele

JorgeBucaran:
@mindplay-dk @JSteunou  @estrattonbailey @pspeter3 @maxholman @dancespiele

pspeter3:
I wonder if it is worth making initial patch and update patch two different methods. Right now the name could be confusing.

[...]

PR #96, opened 2018-02-16
→ Picodom 2.0 → Patchdom

jorgebucaran:
## Summary

The result is a large Double Espresso Steamed Soy Milk Foam!

```html
<body>
  <div id="app"><h1>This heading element will be recycled by default!</h1></div>
</body>
```
```jsx
import { h, patch } from "picodom"

const view = state => (
  <div>
    <h1>{state}</h1>
    <button onclick={() => patch(view(state - 1), root)}>-</button>
    <button onclick={() => patch(view(state + 1), root)}>+</button>
  </div>
)

const root = patch(view(0), document.getElementById("app"))
```

The DOM element passed to <samp>patch</samp> represents the root of your application, **not** a container. The function returns the element and HTML that results from this operation is the same HTML-like markup returned by the view. 

```html
<body>
  <div>
    <h1>0</h1>
    <button>-</button>
    <button>+</button>
  </div>
</body>
```

In addition, Picodom now supports hydration of pre-existing DOM content. This means that instead of throwing away any server-side rendered markup, Picodom will try reuse those elements. For this to work, the markup must be the same markup produced in your view.

```js
// The delay simulate some SSR HTML + HTTP request.
let root

setTimeout(() => {
  root = patch(view(0), document.getElementById("app"))
}, 3000)
```
Notice in the GIF how the heading fades into the page, but when it's replaced by the counter, it's value is simply changed. This means that Picodom reused that heading, instead of creating a new one from scratch and only updated the element's <samp>nodeValue</samp>. Notice how the +/- buttons are animated, however, since their markup was not included (on purpose in order to demonstrate this feature) in the pre-existing HTML. 

## See Also

https://github.com/picodom/picodom/issues/94


lukejacksonn:
If I understand this correctly.. this is so dope

JorgeBucaran:
I was also thinking, that if no DOM element is provided, perhaps we could create the element for you and return it to you, allowing you to append it to the DOM whenever you want.

mindplay-dk:
> Picodom now supports hydration of pre-existing DOM content. This means that instead of throwing away any server-side rendered markup, Picodom will try reuse those elements.

Won't this cause problems with third-party vanilla JS widgets?

If I were to use a third-party lib to create a date-picker or cropping widget using `oncreate`, won't calls to the patch-function wipe out those elements?

And if not, doesn't that mean it also won't wipe out elements that *should* be removed, e.g. if the number of items in a list is lower than it was when it was rendered on the server-side?

The problem with allowing reuse of *any* existing elements you encounter in the DOM, is you can't know if existing elements belong to a rendered view or were created dynamically somehow.

> if no DOM element is provided, perhaps we could create the element for you and return it to you, allowing you to append it to the DOM whenever you want

I don't think that would work for anything that has input/selection/focus state?

JorgeBucaran:
> @mindplay-dk Won't this cause problems with third-party vanilla JS widgets?

No, because you provide the DOM element, we are not guessing it. This works very similarly if not exactly as Preact's render <samp>replaceElement</samp> argument.

> I don't think that would work for anything that has input/selection/focus state?

An example should help clarify this confusion.

```js
import { h, patch } from "picodom"
const element = patch(h("h1", {}, "Hello"))
...
document.body.appendChild(element)
```

The alternative would be to append to <samp>document.body</samp> by default, which I prefer to avoid.

mindplay-dk:
> No, because you give us the DOM element

I can give you the *top* element only - the algorithm then decides what to do with subtrees.

It [does](https://jsfiddle.net/mindplay/ydg0ud6j/) erase any foreign nodes - which is to be expected, how else could you guarantee that a client-side call would actually bring the document up to it's current state. So far so good.

But it [will](https://jsfiddle.net/mindplay/3rga9e98/) then necessarily also destroy foreign nodes created during `oncreate`, since there's no way to tell the difference. (Let that example run for about five seconds, it'll do another update, you'll see the third-party JS widget gets destroyed.)

When you use the DOM as the "source of truth" for the current state of the view, you get these kinds of problems - Picodom 1 didn't reverse translate DOM nodes into VDom nodes, but now it does.

If you don't flag nodes you created yourself, you can't know which DOM state is relevant to an update - this is what I was trying to address in my implementation by injecting the last known VDom state into the DOM, in the hopes that this would also allow for deferred removals, which I also don't think Picodom 2 will presently? If another update is triggered while a not-yet-removed DOM node is still in the DOM, this would be seen as the current state, and it would either add/remove/move the DOM node, which really should be left where it is, would it not?

My hope with my own implementation was that it would only touch DOM nodes that it was either told directly to touch (the top level target node) or had itself flagged during a previous update so it could recognize the node as being a "managed" node. The idea was that DOM nodes lingering during deferred removals, as well as DOM nodes created and injected during `oncreate` would be unaffected.

Note that Preact's top-level `render` function accepts a third argument, the `replaceNode` flag, which allows you to *force* it to replace an entire subtree, as you would if the document had already been rendered from the server-side. I understand you're trying to optimize by recycling the DOM nodes that were created already when the document was loaded, but I think you'd have to at least internally know whether you're forcing the DOM into a specific state, or patching existing managed DOM nodes, as it makes a difference when you have unmanaged nodes that shouldn't be touched.

JorgeBucaran:
> @mindplay-dk But it will then necessarily also destroy foreign nodes created during oncreate...

You can easily avoid foreign nodes to be destroyed if you plan your markup more carefully. Sorry, but your example was only [set up to fail](https://en.wikipedia.org/wiki/Setting_up_to_fail).

```js
let update = model => (
  <div>
    <h4>Hello Here</h4>
    <div>
      <img
        src="http://farm4.staticflickr.com/3095/3502308202_80420dda73_z.jpg"
        oncreate={img => new Cropper(img)}
      />
    </div>
    {model.there && <h4>Hello There</h4>}
  </div>
)
```

> @mindplay-dk I understand you're trying to optimize by recycling the DOM nodes that were created already when the document was loaded...

Hydration is a one-time only operation. If you want to opt-out call <samp>patch</samp> **without** a DOM element. By doing so, a new element will be created and returned to you.

```js
const element = patch(<h1>Hello</h1>)
// 
// ...
//
document.body.appendChild(element)
```

> ...hopes that this would also allow for deferred removals ...

I am **not** trying to solve this problem at the moment, but I may try when I actually rewrite the entire diff algo (in time for Hyperapp 2.0).

mindplay-dk:
> You can easily avoid foreign nodes being destroyed if you plan your markup more carefully

You mean wrap things in extra elements, which is what I mean when I say that HTML support should be universal - I don't want the framework to dictate the kind of markup I can/may use.

I don't think I'm trying to make things unnecessarily difficult here. I just want to be sure that, when my HTML/CSS guy delivers a finished UI design, I can move it into JSX and build a render-function easily, without first having to "fix" the markup and adjust the CSS - and possibly break it, and/or break things for other developers working on other parts of the UI that depend on the same CSS.

> If you want to completely opt-out call patch without a DOM element. By doing so, a new element will be created and returned to you.

😄👍

dancespiele:
Remember to change this in the index.d.ts file before to merge the pull request
```typescript
export function patch(
  parent: Element,
  oldNode: VNode | null,
  newNode: VNode
): Element
```
to
```typescript
export function patch(
  node: VNode,
  parent: Element | null
): Element
```

mindplay-dk:
Note that the second argument is no longer the `parent`, it's the `target` element:

```ts
export function patch(
  node: VNode,
  target: Element | null
): Element
```

PR #98, opened 2018-02-26
2.0

dancespiele:


JorgeBucaran:
Thanks, @dancespiele. Did you take these tests from Hyperapp? Since the lifecycle events are essentially the same as in Hyperapp, I want to copy the tests from Hyperapp to avoid having to maintain two different implementation of the tests.

dancespiele:
the unique thing that I changed for adapting to the new patch was `beforeEach`
```javascript
beforeEach(() => {
  document.body.innerHTML= "";
  const el = document.createElement('div');
  element = document.body.appendChild(el);
})
```
the tests are the same
If you find the way that the tests works with this `beforeEach`
```javascript
beforeEach(() => {
  document.body.innerHTML = ""
})
```
please let me know because I couldn't

jorgebucaran:
Thanks, but the tests are fixed now. 👋

PR #100, opened 2018-03-03
add null possibility to patch types

dancespiele:
I need this if not I get error when I put this:
`patch(node, document.getElementById(this.root));`
because typescript consider that maybe I can get null from `document.getElementById(this.root)`
then throw compatibility types error


jorgebucaran:
Perfect, thanks! 🎉

PR #104, opened 2018-03-06
add proxyFunction types to patch

dancespiele:


jorgebucaran:
Didn't make it to the cut!

PR #105, opened 2018-03-07
fix h types

dancespiele:
with the type which there is now is not compatible with this:

```javascript
{
  nodeName: "div",
  attributes: {},
  children: [
    {
      nodeName: "h1",
      attributes: {},
      children: 0
    },
    {
      nodeName: "button",
      attributes: { ... },
      children: "-"
    },
    {
      nodeName: "button",
      attributes: { ... },
      children: "+"
    }
  ]
}
```
to make it compatible is required to change the h types to:
```javascript
export type Children = VNode | string | number | null;

export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes,
  children?: Array<Children> | Children,
): VNode<Attributes>
```
instead of
```javascript
export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes,
  ...children: Array<VNode | string | number | null>
): VNode<Attributes>
```
because this `...children: Array<VNode | string | number | null>` is translated to:
```javascript
{
  nodeName: "div",
  attributes: {},
    {
      nodeName: "h1",
      attributes: {},
      children: 0
    },
    {
      nodeName: "button",
      attributes: { ... },
      children: "-"
    },
    {
      nodeName: "button",
      attributes: { ... },
      children: "+"
    }
}
```
therefore it is not possible to put an array after the attributes

### DIFF
--- a/ultradom.d.ts
+++ b/ultradom.d.ts
@@ -1,6 +1,5 @@
 export as namespace ultradom
 
-export type Children = VNode | string | number | null;
 /** The virtual DOM representation of an Element. */
 export interface VNode<Attributes = {}> {
   nodeName: string
@@ -16,7 +15,7 @@ export interface Component<Attributes = {}> {
 export function h<Attributes>(
   nodeName: Component<Attributes> | string,
   attributes?: Attributes | null,
-  children?: Array<Children> | Children,
+  ...children: any[],
 ): VNode<Attributes>
 
 /**
--- a/ultradom.d.ts
+++ b/ultradom.d.ts
@@ -1,5 +1,6 @@
 export as namespace ultradom
 
+export type Children = VNode | string | number | null;
 /** The virtual DOM representation of an Element. */
 export interface VNode<Attributes = {}> {
   nodeName: string
@@ -15,7 +16,7 @@ export interface Component<Attributes = {}> {
 export function h<Attributes>(
   nodeName: Component<Attributes> | string,
   attributes?: Attributes | null,
-  ...children: any[],
+  ...children: Array<Children | Children[]>,
 ): VNode<Attributes>
 
 /**
--- a/ultradom.d.ts
+++ b/ultradom.d.ts
@@ -18,7 +18,6 @@ export function h<Attributes>(
   children?: Array<VNode | string | number | null>
 ): VNode<Attributes>
 
-
 /**
  * Patch a DOM element to match the supplied virtual DOM representation. If no element is given, create and return a new DOM element.
  *
--- a/ultradom.d.ts
+++ b/ultradom.d.ts
@@ -14,9 +14,10 @@ export interface Component<Attributes = {}> {
 
 export function h<Attributes>(
   nodeName: Component<Attributes> | string,
-  attributes?: Attributes,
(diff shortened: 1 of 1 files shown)


jorgebucaran:
@dancespiele Good, but does the following usage?

```js
h(nodeName, attributes, child1, child2, child3)
```

dancespiele:
@jorgebucaran  with this
``` 
export type Children = VNode | string | number | null;

export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes | null,
  children?: Array<VNode | string | number | null>
): VNode<Attributes>
```
allow does this:
h(nodeName, attributes, child1, child2, child3)
or:
h(nodeName, attributes, [child1, child2, child3])
and more possibilities

jorgebucaran:
@dancespiele Okay, that's interesting. I don't know understand how <samp>h(nodeName, attributes, child1, child2, child3)</samp> can be supported with these types, but I'll merge now and figure it out later. I love these types. 👍

dancespiele:
you right @jorgebucaran , the unique way to make compatible this 
`h(nodeName, attributes, child1, child2, child3)`
or this:
`h(nodeName, attributes, [vnode, vnode, vnode], child2, ...)`
```typescript
export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes | null,
  ...children: any[],
): VNode<Attributes>
```

jorgebucaran:
@dancespiele The reason `h(nodeName, attributes, child1, child2, ...)` exists is only to support JSX. Is there a way to make JSX not complain with your original type proposal?

dancespiele:
I saw that this is the unique that I now which JSX not complain
```typescript
export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes | null,
  ...children: any[],
): VNode<Attributes>
```

dancespiele:
My last updated changed to this 
```typescript
export type Children = VNode | string | number | null;
export function h<Attributes>(
  nodeName: Component<Attributes> | string,
  attributes?: Attributes | null,
  ...children: Array<Children | Children[]>,
): VNode<Attributes>
```
I explain that. When you write a [Arguments object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) typescript forces to write an array type but doesn't means that the type is an array, it means that every argument could be with the types that is inside of the array for example `...children: Array<VNode | string | number | null>;` means that every children argument can be  `VNode | string | number | null` but not this `[VNode | string | number | null]` the second way has to be typed like this ..children: Array<Array<VNode | string | number | null>> but we want that the children type has the two possibilities then you need to type like this `Array<Array<VNode | string | number | null> | VNode | string | number | null>` and to make it more legible it has to be like this
```typescript
export type Children = VNode | string | number | null;

...children: Array<Children | Children[]>,
```

about this
```typescript
attributes?: Attributes | null,
```
`undefined` is not the same that `null` then you need specify that otherwise wouldn't possible to do this:

```typescript
h("div", null, [...])
```

jorgebucaran:
Thanks, @dancespiele! 🎉

PR #111, opened 2018-04-17
Goodbye patch, hello render.

jorgebucaran:
Rename patch→render and use super simple signature.
    
There is no longer a way to patch a specific element and recycling has been removed. 100% client-side apps.
    
The render function is essentially an overly simplified ReactDOM.render, without component lifecycle, based on Hyperapp's virtual DOM.
    
Mounting takes over the supplied container so make sure your container is empty. 😉 

### Hello World

```jsx
import { h, render } from "ultradom"

render(<h1>Hello, world!</h1>, document.body)
```

### Clock

```jsx
import { h, render } from "ultradom"

const tick = () =>
  render(
    <div>
      <h1>Hello, world!</h1>
      <h2>It is {new Date().toLocaleTimeString()}.</h2>
    </div>,
    document.body
  )

setInterval(tick, 1000)
```

### Counter

```jsx
import { h, render } from "ultradom"

const app = state =>
  render(
    <main>
      <h1>{state}</h1>
      <button onclick={() => app(state - 1)}>-</button>
      <button onclick={() => app(state + 1)}>+</button>
    </main>,
    document.body
  )

app(0) // Start!
```

### JSX Components

```jsx
import * as ultradom from "ultradom"

const h = (name, attributes, ...children) => {
  return typeof name === "function"
    ? name(attributes, children)
    : ultradom.h(name, attributes, children)
}

const Clock = () => (
  <div>
    <h1>Hello World!</h1>
    <h2>The time is: {new Date().toLocaleTimeString()}</h2>
  </div>
)

setInterval(() => ultradom.render(<Clock />, document.body), 1000)
```

#### `createStore.js`

```jsx
export const createStore = (state = {}) => {
  let subscribers = []
  const store = {
    getState: () => state,
    dispatch: reducer => {
      state = reducer(state)
      subscribers.map(s => s(store))
    },
    subscribe: s => {
      subscribers.push(s)
      return () => {
        subscribers = subscribers.filter(sub => sub !== s)
      }
    }
  }
  return store
}
```

#### `app.js`
```jsx
import { h, render } from "ultradom"

export const app = store =>
  render(
    <main>
      <h1>{store.getState()}</h1>
      <button onclick={() => store.dispatch(state => state - 1)}>-</button>
      <button onclick={() => store.dispatch(state => state + 1)}>+</button>
      <button
        onclick={({ dispatch }) => {
          setTimeout(() => {
            store.dispatch(state => state + 1)
          }, 2000)
        }}
      >
        + (1s delay)
      </button>
    </main>,
    document.body
  )
```

#### `index.js`

```jsx
import { app } from "./app"
import { createStore } from "./createStore"

const store = createStore(0)
store.subscribe(app)
app(store)
```


cjh9:
> There is no longer a way to patch a specific element

This is not good for my case, I would like to be able to calculate a part of the vdom-tree, diff and patch from a specific element, not recalculate the whole app at each render...

jorgebucaran:
@cjh9 Bummer.

cjh9:
Yes, that's one way to solve not doing diffing too many nodes. The other way is attaching a `shouldComponentUpdate` function. Maybe both the render-api and patch-api could be supported?

mindplay-dk:
> Mounting takes over the supplied container so make sure your container is empty

Looks like we have the 1:1 single parent/child relationship now forced again? didn't we address that?

Being able to mount/patch a single node directly is pretty essential if anyone's going to attempt to implement a component pattern with this library.

mindplay-dk:
This is not only about component patterns though - we've discussed this at length, and the main problem still is that this forces you to use very specific markup structure, which can make it difficult to move from an HTML/CSS prototype to a working app.

For example, if your CSS guys deliver an HTML prototype with BEM-style CSS, you may have to start picking apart their BEM blocks, adding arbitrary wrapper-elements and (in particular) changing nested CSS selectors.

This being a library and not a framework, it's particularly bad that you can't use arbitrary markup structure according to your own requirements and design.

You also removed the ability to pass `null` for the element and get a new node back, so I can no longer implement the `mount()` function I used to create. As per [this example](https://jsfiddle.net/mindplay/6yev75o0/), I now have to add an arbitrary dummy wrapper-node just for positioning, before I can render anything at all.

If this is how the `render()` function is going to work from now on, I'd like to suggest you expose the `patch()` function as well, which will provide much more flexibility. Yes, calling it is more verbose - but, again, this being a library and not a framework, you will most likely build your own layer, so who cares how verbose that one line of code is?

From where I'm sitting, the `render()` function is seriously in the way of the low-level functionality in the library that I'm interested in.

I might even go so far as to suggest removing the `render()` function altogether, because those couple of lines are not where the utility of this package lies - it's abstracting away the details and thereby much of the control we'd have with `patch()`, and I think that's a real problem.

You removed functional components as well, so you're obviously heading in the direction where this is more of a library and less of a micro-framework, not so much intended to be useful without a layer on top of it.

It kind of seems like that `render()` function belongs in `hyperapp`?

jorgebucaran:
@mindplay-dk React doesn't have a problem with this. Preact has `render(node, container, element)`, but their goals are different to mine.

<samp>ultradom.h</samp> flattens arrays, so the only annoying thing is that the root element is nested inside a dummy div. This problem will go away when fragments are implemented https://github.com/jorgebucaran/ultradom/issues/71.

mindplay-dk:
My point remains valid: the `render()` function walls off something potentially useful - even if it's not useful to your use-case (hyperapp) there's no reason to obstruct others from accessing it?

This being a library, it should expose the functionality of the library rather than abstracting and hiding it - it's for frameworks to decide the degree of abstraction.

Otherwise, I guess, expect people to fork rather than import and use your library, which is what I would have to do to explore the kind of use-cases I'm interested in. But that seems kind of silly, since I could have just forked your framework then - sorta proves my point doesn't it? Why is it even a library if I can't access the library functionality and end up having to treat it like a framework? hmm.

jorgebucaran:
The ability to target a specific element and SSR recycling combined allows a pattern that emulates stateful components. I think this is misuse. You want to use the right tool for the right job and stateful components is a solved problem — just use React or Preact for that.

PR #113, opened 2018-05-05
add types file to package.json so typescript can find it

tscholl2:
This PR adds the `types` field to the package.json file following the official recommendations:

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

I had trouble with typescript not finding the global `JSX` type definitions defined in `ultradom.d.ts` until I did this.


jorgebucaran:
Thanks, @tscholl2! 🎉

PR #124, opened 2018-06-28
Fixed JSX example in README.

rbiggs:
1. JSX example was missing children parameter, causing browser to throw error of undefined variable.


jorgebucaran:
Fixed https://github.com/jorgebucaran/superfine#components. 👋

PR #142, opened 2018-07-09
Update superfine.d.ts

tscholl2:
Looks like v6 renamed "render" -> "patch". This should update the typings accordingly.


jorgebucaran:
Whoops! Thank you, @tscholl2! 👍

PR #144, opened 2018-07-11
update typings

tscholl2:
This updates the type definitions with the latest changes.

One small change, I made `props` not optional on the interface because everywhere where `createVNode` is called, the props is defined. I don't know if that is what you intended semantically though.

### DIFF
--- a/superfine.d.ts
+++ b/superfine.d.ts
@@ -6,10 +6,12 @@ export type Children = VNode | string | number | null
  * The virtual DOM representation of an Element.
  */
 export interface VNode<Props = {}> {
-  name: string
-  props?: Props
-  children: Array<VNode>
-  key: string
+  name: string,
+  props: Props,
+  children: Array<VNode>,
+  element: Element | undefined,
+  key: string | null,
+  type: number
 }
 
 /**
--- a/superfine.d.ts
+++ b/superfine.d.ts
@@ -9,7 +9,7 @@ export interface VNode<Props = {}> {
   name: string,
   props: Props,
   children: Array<VNode>,
-  element: Element | undefined,
+  element: Element | null,
   key: string | null,
   type: number
 }


superfine.d.ts
```diff
@@ -6,10 +6,12 @@ export type Children = VNode | string | number | null
...
  */
 export interface VNode<Props = {}> {
-  name: string
-  props?: Props
-  children: Array<VNode>
-  key: string
+  name: string,
+  props: Props,
+  children: Array<VNode>,
+  element: Element | undefined,
```
jorgebucaran: If the vnode is created using `h`, then, I think `element` can only be null.

superfine.d.ts
```diff
@@ -6,10 +6,12 @@ export type Children = VNode | string | number | null
  * The virtual DOM representation of an Element.
  */
 export interface VNode<Props = {}> {
-  name: string
-  props?: Props
```
jorgebucaran: @tscholl2 Does Props? means Props are optional? If that's so, then it shouldn't be there.

tscholl2: `props?: Props` means that the `props` property may or may not exist on a `VNode`, and if it does it exist, its type will be `Props`.

The reason I got rid of the `?` is that when I looked at the code, I thought that if a `VNode` is created through the `h` function, the `props` property is always set on `VNode`, and moreover, it is not null/undefined. I got that idea from this line:

https://github.com/jorgebucaran/superfine/blob/master/src/index.js#L422

Let me know if I am misunderstanding the code, or if this does not match the intended use of `VNode`. For example, if you intend other people to create `VNode`s through their own process and you want to convey that they do not need to set `props` on a `VNode` for the patch function to work (I didn't check to see if `patch` assumes the existence of this property).

jorgebucaran: @tscholl2 I misunderstood the diff (yes, now that's a thing). I thought you added a `?`, but you _removed it_, which is 100% correct. Sorry! 😅

Issue #134, opened 2018-07-04
What's coming next & Superfine v. Hyperapp


jorgebucaran:
Here is a mixed list of current and upcoming Superfine features and changes. Some of these may not be available to Hyperapp or vice-versa. I'm planning to work through these after Hyperapp V2 is out.

- Typescript rewrite.
    -  Use more ES6 features, e.g., Map, arrow functions, destructuring.
- Superfine will continue to support lifecycle events as a pragmatic solution for enter/exit animations, encapsulation of state, global event listeners, etc. See  https://github.com/hyperapp/hyperapp/issues/717 for more uses and why they're going to be removed from Hyperapp.
- Lazy views. See: https://github.com/hyperapp/hyperapp/issues/721
    - Lazy views allows you to declare a part of your VDOM should be re-rendered only when some data changes. This features is equivalent to Elm's [Html.lazy](http://package.elm-lang.org/packages/elm-lang/html/2.0.0/Html-Lazy).
- Expose `eventProxy`. 
    - Intercepting DOM events allows Hyperapp to couple VDOM and state management. Another use could be auto-redrawing. What else could you use it for? 
- ~~JSX fragments (vnodes of type array) #71.~~
- ~~Explore **class components** (not available to Hyperapp)~~
- Shadow DOM?

---

rbiggs:
Was wondering if you're familiar with Raj: https://github.com/andrejewski/raj. Kinda similar goals to Hyperapp -- Elm Architecture for JavaScript.

jorgebucaran:
Yep.

brandonros:
As it stands, what does Superfine provide that Hyperapp 1.0 does not and vice versa?

rbiggs:
@brandonros Superfine is just a Virtual DOM. You could use it with functional components and JSX for a bare bones fill-in for P/React. It's only 1KB.

Hyperapp has a Virtual DOM but also provides state management inspired by the Elm Architecture. If you're using Superfine and need state management, you'll need to include Redux, Mobx or write your own solution. If you use Hyperapp, you get everything you need.

rbiggs:
@jorgebucaran, you mention a possible TypeScript rewrite for Superfine. What if you just stick with JavaScript and use JSDoc comments to define the types? If you're using VSCode or Webstore, you'll get type checking and intellisense right away. For VSCode you just add the following to you user settings:
```javascript
"javascript.implicitProjectConfig.checkJs": true
```
[...]

PR #155, opened 2018-12-05
migrate to terser, sets foundation for typescript migration

### LINKED ISSUE #150: Using Terser to minify
FYI I've had a lot of luck with [Switzerland](https://github.com/Wildhoney/Switzerland) in minifying JS modules that's still usable in the browser with [Terser](https://github.com/fabiosantoscode/terser).

The way I go about it is to offer both side-by-side:

* https://cdn.jsdelivr.net/npm/switzerland@3.3.0/es/production/index.js
* https://cdn.jsdelivr.net/npm/switzerland@3.3.0/es/development/index.js

Have you thought about minification? Would you like me to try and implement it in Superfine?

osdevisnot:
**What this PR does**
- [x] moves minify from now unmaintained uglify to terser, closes #150 
- [x] sets foundation for typescript migration as mentioned in #134 
- [x] adds prettier support for consistent formatting
- [x] use more ES6/Typescript features
  - [x] arrow functions
  - [x] const & let

**Future Work**
- [ ] switch to typescript strict mode
- [ ] Use more ES6/Typescript features (Map, destructuring, etc.)

**Todo before merging this PR**
- [ ] match typings for public functions
- [ ] squash all commits into one (or at least fewer) to reduce noise.

Please note, this initial PR attempts to achieve above features with least possible changes to src. Main objective for this PR is to collect feedback from community.

Thoughts / Opinions ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,10 @@ export const ImageGallery = images =>
   )
 ```
 
+### Top-Level Nodes
+
+Keys are not registered on the top-level node of your view. If you are toggling the top-level view, and you must use keys, wrap them in an unchanging node.
+
 ## Lifecycle Events
 
 You can be notified when elements managed by the virtual DOM are created, updated or removed via lifecycle events. Use them for animation, wrapping third party libraries, cleaning up resources, etc.


osdevisnot:
@mindplay-dk Thank you for checking. So far I tried using `git mv` (see first commit in this PR) then apply change. I also tried a fresh commit with all changes applied and files moved using `git mv` (see develop branch on my fork).

However, nothing seems to provide a diff in github. The history on individual files, is now correctly preserved (with `git mv`), for eg, here is what my webstorm tells me for history of main file:

image

I will try more tomorrow to get github show correct diff, but their algorithm appears to have few hardcoded limits (see [diffcore-renam.c](https://git.kernel.org/pub/scm/git/git.git/tree/diffcore-rename.c?id=HEAD)

Funny enough, when I navigate individual commits, I am able to see correct diff. Check this for example -- https://github.com/jorgebucaran/superfine/pull/155/commits/88e44876be9c368795ffda6ffa7c15b57e901601

mindplay-dk:
> Funny enough, when I navigate individual commits, I am able to see correct diff.

Yeah, huh. Not your bad then. Probably should file a bug report with GitHub?

Now that I can see the diff - I see you've added semi-colons to the entire codebase; I don't personally have any strong feelings one way or the other, but that's a style-change, unrelated to the changes you're proposing, and it's making the diff extremely noisy and hard to read.

A code style change probably belongs in a separate proposal. Which I'm fairly sure @jorgebucaran would reject 😉

osdevisnot:
I somehow overlooked code style changes. Changed from semi to no-semi and from single quote to double quote in my latest commits.

Also, it seems github magically started showing the PR diff as well : 
image

osdevisnot:
Also, this is how generated d.ts file looks so far
```
declare global {
    namespace JSX {
        interface Element extends VNode<any> {
        }
        interface IntrinsicElements {
            [elemName: string]: any;
        }
    }
}
    name: string;
    props: Props;
    children: Array<VNode>;
    element: Element | null;
    key: string | null;
    type: number;
}
export declare type Children = VNode | string | number | null;
export declare const recycle: (container: any) => {
    name: any;
    props: any;
    children: any;
    element: any;
    key: any;
    type: any;
};
/**
 * Render a virtual DOM node into a DOM element container.
 * @param lastNode The last virtual DOM node.
 * @param nextNode The next virtual DOM node.
 * @param container A DOM element where the new virtual DOM will be rendered.
 */
export declare const patch: (lastNode: VNode<{}>, nextNode: VNode<{}>, container: Element) => VNode<{}>;
export declare const h: (name: any, props?: any, ...rest: any[]) => any;
```

Next up, figure how to correctly type `h` (current approach causes error when we try to access children in `h`)

mindplay-dk:
> Also, it seems github magically started showing the PR diff as well

Likely because the files are now similar enough to look like the same file :-)

jorgebucaran:
@osdevisnot I agree, migrating to terser and migrating to TypeScript are different things. Could this PR only take care of the first?

mindplay-dk:
Yeah, the switch to terser makes sense now - while the TypeScript switch is probably a much longer process.

A separate PR for the switch to terser would make a lot of sense.

osdevisnot:
@mindplay-dk @jorgebucaran makes sense. I will close this PR and create 2 separate ones for separation of concerns.

PR #156, opened 2018-12-20
closes #150 Move to terser

### LINKED ISSUE #150: Using Terser to minify

osdevisnot:
closes #150 Move to Terser


.babelrc
```diff
@@ -0,0 +1,13 @@
...
+    "react",
+    "stage-0"
+  ],
+  "plugins": [
+    "transform-class-properties",
+    "transform-decorators",
+    "transform-react-constant-elements",
+    "transform-react-inline-elements"
+  ]
+}
```
osdevisnot: @jorgebucaran @mindplay-dk It seems the terser CLI adds default babelrc with above presets and plugins. However having an empty babelrc also works for our purpose. I am more inclined to keep this file but would be interested in hearing your thoughts.

jorgebucaran: Assuming that's the case, does anyone know why terser adds it?

osdevisnot: On further investigation, terser package is ok. But, babel-env has a postinstall script https://github.com/epsitec-sa/babel-env/blob/master/package.json#L9 which creates a default babelrc. This feature is also documented in README.md for this package https://github.com/epsitec-sa/babel-env#babel-env

image

osdevisnot: Currently we need it for jest. Without env, jest would complain for not finding   env preset. Alternative would be to add these packages:
```
yarn add --dev babel-jest babel-core@^7.0.0-bridge.0 @babel/core regenerator-runtime
```

jorgebucaran: @osdevisnot Yep, thanks for clarifying it.

jorgebucaran: Right, thanks for clarifying that.

osdevisnot:
@jorgebucaran I had to move from babel-env to babel-preset-env to remove default generated `.babelrc` and be able to use the babel config from `package.json`.

jorgebucaran:
@osdevisnot Thanks! 🎉

jorgebucaran:
@osdevisnot Thanks! 🎉  


One problem though. Any idea why jest fails when I run npm test locally?

```
Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
```

jorgebucaran:
Yeah, I did. I'd never seen that error before. I'll keep trying.

osdevisnot:
And you are on at least node 8 or above?

jorgebucaran:
@osdevisnot Ah, yeah, that was it! I was using Node 7 for some reason. 😆

osdevisnot:
Glad it worked for you 👍👍

PR #158, opened 2018-12-21
typescript foundation

osdevisnot:
Tasks Completed:
- [x] Create type definitions for custom matcher used in tests
- [x] Configure tests to use `ts-jest` over `jest` and babel
- [x] Remove babel config from package.json
- [x] Refactor builds to use rollup with `rollup.config.js` over commands in scripts section of `package.json`
- [x] Repoint `main`, `module` and `typyings` fields in `package.json` to dist
- [x] Add new packages to `package.json`
- [x] Remove existing `.d.ts` file as this will now be auto generated.

Tasks Pending:
- [ ] Match type definitions for public functions
- [ ] add better types for internal functions
- [ ] use more typescript features (rest/spread, arrow functions, Map, etc.)
- [ ] switch to ts `strict` mode


osdevisnot:
@jorgebucaran @mindplay-dk this is continuation of TS porting efforts that were being discussed in #155. Currently this is on a branch in my-fork and the PR is against master of this repo.

src/superfine.ts
```diff
@@ -411,14 +442,21 @@ export var patch = function(lastNode, nextNode, container) {
...
-export var h = function(name, props) {
+/**
+ * create a new vdom node. vdom is a description of what DOM should look like using a tree of virtual nodes.
+ * @param name name of an element or function that returns vdom node
+ * @param props HTML props, SVG props, DOM events, lifecycle events and keys
+ * @param rest child nodes for an element
+ */
+export var h = function(
+  name: string | Function,
+  props?: any,
```
osdevisnot: @mindplay-dk existing definition for h are open to Props being a type. This works ok when declaring TS definitions separately. Any pointers on how I can achieve that using TS definitions in file?

mindplay-dk: The only difference as far as I can figure, is the `h` function will have a body - other than that, you should be able to just copy and paste from the original `.d.ts`?

You'll want to exclude the `.d.ts` from the build (as you do already) and move to emitting a generated `.d.ts` instead - once you have all the type declarations moved over, the emitted `.d.ts` should be very close to the hand-written one we have now.

(I'm not sure if that's what you were asking?)

jorgebucaran:
The codebase has changed quite a bit. If this is going to happen, we might as well start from scratch.

PR #161, opened 2019-01-15
Allow the passing of HTMLElement as the node name

Wildhoney:
I'm experimenting with an approach where you can pass the `HTMLElement` to be used for the tree itself, rather than letting `superfine` create the element itself to then depend on the `oncreate` callback to gather references.

For instance:

```javascript
const todoInput = document.createElement('input');

h('div', {}, [
    h(todoInput, { type: 'text', onchange: render }),
    h('div', {}, `Valid: ${todoInput.validity.valid}`)
])
```

Would be super useful as I'd be able to use `todoInput` as it is without any `null` checking.

The alternative being to let `superfine` create a random `HTMLElement` itself and use `oncreate` to re-render the tree &ndash; which causes three *problems*: `null` checking to ensure we have `todoInput`, multiple passes of DOM reconciliation (`oncreate` × node count), and a little boilerplate for the `oncreate` itself:

```javascript
h('div', {}, [
    h('input', { type: 'text', oncreate: node => (todoInput = node, render()) }),
    todoInput.validity.valid && h('div', {}, `Valid: ${todoInput.validity.valid}`)
])
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,12 +82,11 @@ var updateProperty = function(element, name, lastValue, nextValue, isSvg) {
 }
 
 var createElement = function(node, lifecycle, isSvg) {
-  var element =
-    node.type === TEXT_NODE
-      ? document.createTextNode(node.name)
-      : (isSvg = isSvg || node.name === "svg")
-        ? document.createElementNS(SVG_NS, node.name)
-        : document.createElement(node.name)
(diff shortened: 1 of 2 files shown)


jorgebucaran:
Without lifecycle events in Superfine 7 this is not really as useful. Thanks anyway!

PR #165, opened 2019-06-07
Fix README.md

sVaterlaus:
Fix babel plugin name from "transform-react-jsx" to "@babel/plugin-transform-react-jsx".

("transform-react-jsx" was giving me weird errors until I changed it)


jorgebucaran:
Good catch, thanks! @sVaterlaus 👋

Issue #172, opened 2019-08-27
V7 and Document Fragments


osdevisnot:
I really love new simplified `patch` usage and the simplified algorithm. 

However, it seems we assume that the `node.parentNode` always exists in https://github.com/jorgebucaran/superfine/blob/master/src/index.js#L265

This works for the most part, but fails when the provided node is a [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment), since DocumentFragment won't have a `parentNode`.

The `patch` in this case errors out with:

![image](https://user-images.githubusercontent.com/802242/63737610-e01af700-c83b-11e9-84bb-8c7ee00ffc55.png)

To add more context, I have a web-components wrapper based on `Superfine`, which was invoking the `patch` in v6 as https://github.com/osdevisnot/supertag/blob/master/src/supertag.ts#L94

With V7, I believe the correct usage would simply be something like:
```js
this[ROOT] = patch(this[ROOT], this.render());
```

But this fails with above error since `this[ROOT]` is a DocumentFragment created using 
```js
this[ROOT] = this.attachShadow({ mode: 'open' });
```

Any pointers on how we can use the V7 patch with DocumentFragments? Or am I missing something?

jorgebucaran:
@osdevisnot What do you suggest?

iosio:
I've been wrestling with this myself. Attempting to use superfine to render inside the shadowRoot of a web component, and utilize some form of a fragment as the root element. Time well spent though because this is by far the smallest and most awesome vdom ever. I was just about to submit an issue or somehow reach out. 

I got a <Fragment/> element working and am able to patch and re-patch subsequent updates reliably. Just kind of a hack and a pain to try and test.

demo:
https://github.com/iosio/x/blob/master/demo/index.js

vdom 
https://github.com/iosio/x/blob/master/src/superfine_modified%26annotated.js#L136

web component base
https://github.com/iosio/x/blob/51f6e2c558507695026ee284727605de94729802/src/x.js#L56

iosio:
I got a Fragment element working and am able to patch and re-patch subsequent updates reliably into the shadowRoot of a web component. Just kind of a hack. Would be awesome to find a better solution.

demo:
https://github.com/iosio/x/blob/master/demo/index.js

vdom - modified superfine
https://github.com/iosio/x/blob/master/src/superfine_modified%26annotated.js#L136

web component base - some trickery with conditional mounting point 
[...]

PR #174, opened 2019-09-12
change patch semantics to append mode

osdevisnot:
This is one possible fixe for #172. We change patch semantics to use append to container like in v6

I assume this will break SSR as recycling might need more work. But without tests added to V7 branch, I can't tell this certainly.

I will try to create another PR which will play with replace container mode.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -262,13 +262,17 @@ var recycleNode = function(node) {
 export var patch = function(node, vdom) {
   return (
     ((node = patchNode(
-      node.parentNode,
       node,
-      node.vdom || recycleNode(node),
+      node.children[0],
+      node.vdom,
       vdom
     )).vdom = vdom),
     node
   )
+}
+
+export var recycle = function(container) {
+  return recycleNode(container.children[0])
 }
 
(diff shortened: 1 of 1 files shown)


osdevisnot:
@jorgebucaran I would like to see if this approach would work for SSR scenarios. Do you have WIP tests that I can use. If not, I can try importing relevant tests from V6.

Also, for root node semantics, did we have strong reason to replace vs append? The reason I ask is because it's a bit untraditional to follow the later.

jorgebucaran:
@osdevisnot I don't want to append. I want to replace (recycling if possible).

osdevisnot:
Got it, I'm guessing it brings obvious advantages which I am definitely missing here. If it's not an hassle, do you mind enlightening me on 'why' part of replace vs append?

jorgebucaran:
I want to take over an HTML node and replace it with my own view. My view can be anything I want or it can be the same as existing HTML content in the document to take advantage of Superfine's recycling.

I also want to find a solution to #172, but not by changing Superfine's default patch behavior. What other options do we have?

PR #175, opened 2019-09-12
change patch semantics to append mode only for fragments

osdevisnot:
This is another possible fixe for #172. We change patch semantics to use append to container like in v6 but only for `DocumentFragments`.


jorgebucaran:
@osdevisnot This is the other option we have, I guess.

osdevisnot:
@jorgebucaran yes this is an option too. Let me know if you would like something revised here.

jorgebucaran:
@osdevisnot Not at the moment, I just need time to review this carefully and study its implications. Have you found anything so far?

PR #177, opened 2019-11-22
made objects scoped locally

timonson:


jorgebucaran:
Not a bug: https://github.com/jorgebucaran/superfine/issues/176#issuecomment-557526804

Issue #184, opened 2021-03-28
Support for React-like `ref`s?


mcjazzyfunky:
Wouldn't  it be a nice feature if `superfine` supported `ref` pseudo properties like [React](https://reactjs.org/docs/refs-and-the-dom.html), Preact, Dyo, uhtml etc. do?

```js
const elemRef = { current: null }
// [...]
const output = <some-element ref={elemRef} />
```

Here's a little demo (based on a R&D web-component toy library that uses a `ref`-supporting patched version of `superfine` internally):
https://codesandbox.io/s/counter-demo-r6m32?file=/src/index.tsx

FYI: Here's a naive quick'n'dirty implementation which obviously only provides part of the requested functionality: [LINK](https://github.com/js-works/js-element/blob/7742c11d3a6f6ffff2353fae3cc017f95c4c81a9/src/main/lib/patched-superfine.ts#L26)

jorgebucaran:
What's the fully requested functionality?

mcjazzyfunky:
> What's the fully requested functionality?

I basically meant that:

- In analogy of React, both ref objects `{ current: null }` and ref callbacks `(elem) => {...}` should be supported.
- Also the corresponding ref object should be cleared (=> `elemRef.current = null`) or the ref callback should be called with `null` (=> `elemCallback(null)`) when the `ref` property has changed or the corresponding element has been disconnected.

For the latter please check the following React demo (everything is working fine there ... Preact and Dyo are working accordingly):
https://codesandbox.io/s/ref-demo-b0kmw
... and compare it with that [naive superfine ref support implementation](https://github.com/js-works/js-element/blob/7742c11d3a6f6ffff2353fae3cc017f95c4c81a9/src/main/lib/patched-superfine.ts#L26) that I have mentioned above:
https://codesandbox.io/s/proud-wind-r25v2?file=/src/index.js

By the way: This is also not working with the unmodified [uhtml library](https://github.com/WebReflection/uhtml):
https://codesandbox.io/s/proud-wind-r25v2?file=/src/index.js

mcjazzyfunky:
Sorry, I have mixed up some of the URLs in my previous comment - have fixed it.

jorgebucaran:
We don't really have to support the exact API as React, Preact, since Superfine does not seek compatibility with any of them, but if that's all it takes or even a little more, I'd definitely consider adding this to the project.

Do you want to work on a PR?

mcjazzyfunky:
[...]

PR #185, opened 2021-03-30
Added support for `ref` properties (see #184)

mcjazzyfunky:
@jorgebucaran Could you please review?

### DIFF
--- a/tests/refs.test.js
+++ b/tests/refs.test.js
@@ -96,8 +96,6 @@ export default [
         const render = (vnode) => patch(container.firstChild, vnode)
         const refObject = createRefObject()
         const refCallback = createRefCallback()
-        
-        container.appendChild(document.createElement("span"))
 
         render(h("div", { ref: refObject }))
         yield equal(refObject.current?.tagName, "DIV")
--- a/tests/refs.test.js
+++ b/tests/refs.test.js
@@ -89,7 +89,7 @@ export default [
       }),
     ]),
     t("test behavior of refs", [
-      t("ref objects and ref callback should work properly", function* () {
+      t("ref objects and ref callbacks should work properly", function* () {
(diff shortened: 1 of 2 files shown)


tests/refs.test.js
```diff
@@ -0,0 +1,138 @@
...
+
+  return ret
+}
+
+// === tests =========================================================
+
+export default [
+  t("ref support for superfine", [
+    t("test ref helpers (only needed for testing)", [
+      t("testing createRefObject", function* () {
```
jorgebucaran, line 56: Why not use `async` functions instead?

tests/refs.test.js
```diff
@@ -0,0 +1,138 @@
+import { t as tw, equal } from "twist"
+import { JSDOM } from "jsdom"
+import { h, patch } from "../index.js"
+
+// we enhance twist's original function `t`
```
jorgebucaran, line 5: Why do we need generators to test this?

mcjazzyfunky: I guess this may have been a misunderstanding on my side.
I was of the impression that in `twist` "one test" means "one invocation of a twist assert function", so this refs test file would contain a bit more than 40 tests ... and to be count properly in the summary I had to `yield` each single one of them.
Normally I would say this refs test file contains only three tests (with over 40 assertions).
Please show me a few lines on code how this is to be done with `twist` normally - thanks - and I will fix it.

jorgebucaran: No worries! If we merge this feature, I'll take care of the tests myself. 🙆‍♂️

jorgebucaran:
Do you have an example using `ref` with this patch to animate removing an element from the DOM?

Won't get too far with refs if we can't provide a solution for that.

jorgebucaran:
@mcjazzyfunky

Issue #188, opened 2022-01-13
Discussion about patching shadow-root nodes? Feature request?


whaaaley:
I just want to preface this issue by saying this is just an idea I had. I'm not sure that this belongs in Superfine but I do think it's interesting. I'm not aware of other vdom libraries doing this. If it's not within scope of Superfine I totally understand.

Lately I've been experimenting with using shadow dom to create scopes for styles to avoid an unnecessarily complicated build step to prefix styles. I can do this already with a strategy that looks like this. I create a "root" dom node and using a node reference (`ref`) from the parent I `attachShadow` using that "root" dom node. Then patch the "root" node. Then I can pass styles into the shadowDom using props. The styles inside cannot leak.

```js
import { patch } from 'superfine'

export default function () {
  let ref = null
  const root = document.createElement('div')

  function handler () {
    ref.node.attachShadow({ mode: 'open' }).appendChild(root)
  }

  return function (props, children) {
    const style = props.style

    if (ref === null) {
      requestAnimationFrame(handler)
    }

    patch(root, <div id={props.id}>
      {style != null && <style>{style}</style>}
      {children}
    </div>)

    // eslint-disable-next-line no-return-assign
    return ref = <div></div>
  }
}
```

This works great but there are a few less than ideal things. One of which is that it recreates the root node and forces re-runing the patch function inside this "component" on every update from the parent. Perhaps this isn't actually a problem. Maybe this has no performance impact at all. But, maybe there's an opportunity to optimize?

Another issue is the use of requestAnimationFrame. This is just how it has to happen in order to grab the node reference. So, upon first mount the shadowDom innards are one frame behind. Subsequent patches should be fine and not lag because there's no RAF-ing. Not a big deal but wouldn't be needed if integrated into the lib. 

Anyway, my thought was that superfine could potentially expose a function to create a shadow context that is "updatable" by the parent. It could potentially look something like this. Please don't mind my use of JSX syntax 😅. Or maybe a property similar to `key` would be better. `<div shadow={true}>`

```js
import { patch, shadowRoot } from 'superfine'
import style from './style.css'

const view = <div>
  <h1>I am an application</h1>
  <shadowRoot>
    <style>{style}</style>
    <h1>I am using styles that are inaccessible to the parent</h1>
  </shadowRoot>
</div>

const root = document.createElement('div')
patch(root, view)
```

[...]

PR #189, opened 2022-02-01
Add support for declarative shadow dom

whaaaley:
This addresses my issue #188.

This PR would allow you to write views that create `shadow-roots`. For example when using JSX:

```js
function view () {
  return <div>
    <h1>Heading 1</h1>
    <div>The styles inside this shadow-root below are completely scoped to that node. This pairs great with importing CSS from JavaScript imports.</div>
    <div shadow-root='open'>
      <h1>
        <style>{'h1 { display: flex; padding: 12px; color: white; background: dodgerblue; } div { padding: 0 24px; }'}</style>
        <button onclick={increment}>Increment</button>
        <div>{state.count}</div>
        <button onclick={decrement}>Decrement</button>
      </h1>
    </div>
  </div>
}
```

Where the DOM output would look like: (The root `h1` inside `shadow-root` is required to mount children into it and the parent `div` has to exist so we have a place to mount the `shadow-root` to.)

image

Here's a codepen of the PR in action. <https://codepen.io/dustindowell/pen/rNYeLxg>

If this is not something you're interested in it's no problem! I just think it's really cool to have encapsulation of styles and scripts without having to touch the custom elements api!

To briefly walk over the changes.
  + I added a `slot` variable, because children will need to go into the `shadow-root` instead of the original node if the property `shadow-root` exists.
  + Inside the new condition, these are the minimum steps needed to create a new shadow node. Shadow root elements can only be a few different tags. The list of valid root tags can be found here: <https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow>. However the most common is obviously `<div>`. ~~I defaulted to `div` because I couldn't think of a good way to configure that. This is maybe something that could adjusted.~~ (see changes)
  + Finally, changing the `node.appendChild` to `slot.appendChild` so the children nodes are appended to the correct parent node.

EDIT: ~~Perhaps a solution to configuring the node you attach inside the `shadow-root` would be to not allow "fragments" inside a node which has a `shadow-root` property. I'll think about this and make some changes.~~ (see changes)

EDIT 2: ~~I also didn't account for patching. 😅  So I do have some work to do still.~~ (patching works fine)

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,6 +86,17 @@ var patchNode = (parent, node, oldVNode, newVNode, isSvg) => {
       parent.removeChild(oldVNode.node)
     }
   } else {
+    if (parent.shadowRoot) {
+      patchNode(
+        parent.shadowRoot,
+        parent.shadowRoot.firstChild,
+        oldVNode,
+        newVNode,
+        isSvg
+      )
+      return // idk
+    }
+
     var tmpVKid,
       oldVKid,
       oldKey,
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ var getKey = (vdom) => (vdom == null ? vdom : vdom.key)
 
 var patchProperty = (node, key, oldValue, newValue, isSvg) => {
   if (key === "key") {
-  } else if (key ==="shadow-root") {
+  } else if (key === "shadow-root") {
   } else if (key[0] === "o" && key[1] === "n") {
     if (
       !((node.events || (node.events = {}))[(key = key.slice(2))] = newValue)
--- a/index.js
+++ b/index.js
@@ -77,6 +77,14 @@ var patchNode = (parent, node, oldVNode, newVNode, isSvg) => {
     newVNode.type === TEXT_NODE
   ) {
     if (oldVNode.tag !== newVNode.tag) node.nodeValue = newVNode.tag
+  } else if (parent && parent.shadowRoot) {
+    return patchNode(
+      null,
+      parent.shadowRoot.firstChild,
+      oldVNode,
+      newVNode,
+      isSvg
+    )
   } else if (oldVNode == null || oldVNode.tag !== newVNode.tag) {
     node = parent.insertBefore(
       createNode((newVNode = vdomify(newVNode)), isSvg),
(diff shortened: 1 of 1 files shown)


whaaaley:
After some simple tests, it seems like patch doesn't need to be changed for this to work. Let me know if I'm missing something. Here is an updated codepen that includes subsequent patching: <https://codepen.io/dustindowell/pen/rNYeLxg>

EDIT: Ok there's problems.

whaaaley:
I believe I've fixed everything.

jorgebucaran:
@whaaaley Awesome. How are you supporting JSX in your pen?

whaaaley:
@jorgebucaran I'm using the typescript option in the pen settings then running that through a function based originally on zaceno's jsx pragma that I tweaked a bit. I just pasted it into the pen for experimenting.

whaaaley:
@jorgebucaran I don't know if you're interested in supporting declarative shadow DOM at all. However, last year Chrome added support for declarative shadow DOM syntax. I was unaware of that syntax at the time of this PR. If you're not sold on adding support yet, totally fine, though it would help me out a getting static rendering and SSR working with Superfine views which use shadow DOM. (Isomorphic code and all that)

The syntax Chrome supports.
```html
<div> <!-- this div has a shadow root attached -->
  <template shadowroot='open'> <!-- this template tag gets obliterated by the browser -->
    <h1>hello world</h1> <!-- contents are added to the shadow root -->
    <style>
      h1 { color: red; }
    </style>
  </template>
</div>
```

Here's a codepen showcasing the above syntax.
<https://codepen.io/dustindowell/pen/dymgvNg>

Some details of my changes in my last commit to this PR. I removed what I had previously and did something different. When iterating over children I check for a shadow root template, If that check is true I "abort" appending children and forward the contents of that template to a new `createNode` function. This is the most efficient way I could think of doing this. On that next `createNode` call, I pass in the mode of the shadow root. If it exists, attach a shadowRoot and continue.

jorgebucaran:
Thanks, @whaaaley. I'll have a look at this. 👍

